PR #19, opened 2023-11-20
Fix computation of response length in the general case

odelalleau:
This is a resubmit of #4 after rebasing. ~~Note that I dropped the changelog update (dc0a979c) since the release is still pending.~~


shengyangs:
Is there any particular reason why `\x11` is not supported here? If I modify the code here to include `\x11`, will it incur some problem? https://github.com/NVIDIA/NeMo-Aligner/blob/main/nemo_aligner/models/nlp/gpt/megatron_gpt_ppo_actor.py#L69

odelalleau:
> Is there any particular reason why `\x11` is not supported here? If I modify the code here to include `\x11`, will it incur some problem? https://github.com/NVIDIA/NeMo-Aligner/blob/main/nemo_aligner/models/nlp/gpt/megatron_gpt_ppo_actor.py#L69

Yeah it will break unless you grab this PR. Sorry I had completely forgotten about this PR, I'll need to fully test it to merge it in, but it should work for you.

odelalleau:
> it should work for you.

Just a heads-up that I actually ran into some issues when trying it out, so it may not work just yet. Unfortunately I won’t be able to look into it until Jan. 4th.

shengyangs:
No worries. I managed to find another hacky solution. Yet still tuning RLHF performances. I will talk with you after you are back. Enjoy the holiday.

odelalleau:
I fixed the issue and rebased. Will test it more thoroughly now. The problem was that
```python
self._end_idx = torch.where(started & is_end & (self._end_idx < 0), context_length, self._end_idx)
```
was missing the parentheses around `self._end_idx < 0`

odelalleau:
Un-drafting as I think I have run enough experiments with this code to be confident it works fine:
* With byte tokens (`\x11`)
* With `<extra_id_1>` tokens
* With `<extra_id_1>` as text

nemo_aligner/utils/text_generation_utils.py
```diff
@@ -14,11 +14,60 @@
...
+    def end_of_generation_condition(
+        self, tokens: torch.Tensor, prev: torch.Tensor, eod_id: int, end_strings: List[str]
+    ) -> torch.Tensor:
+        is_end = super().end_of_generation_condition(tokens=tokens, prev=prev, eod_id=eod_id, end_strings=end_strings)
+        assert len(is_end) == len(tokens)
+        if len(tokens) != len(self._context_lengths):
+            raise RuntimeError(
+                "Batch size mismatch: the `context_lengths` tensor provided in the constructor has batch size "
+                f"{len(self._context_lengths)}, while the generated tokens have batch size {len(tokens)}"
+            )
```
gshennvm, line 48: this logic looks to guard against the batch size changing, in what cases can this occur?

nemo_aligner/models/nlp/gpt/megatron_gpt_ppo_actor.py
```diff
@@ -280,19 +262,36 @@ def infer(self, inference_batch):
...
+        strategy = TrackLengthGPTModelTextGenerationStrategy(
+            model=self, context_lengths=prompt_lengths, max_length=self._length_params["max_length"]
+        )
         actor_output = self.generate(
-            inputs=inputs, length_params=self._length_params, sampling_params=self._sampling_params
+            inputs=inputs, length_params=self._length_params, sampling_params=self._sampling_params, strategy=strategy
         )
 
+        response_lengths = None
+        if parallel_state.is_pipeline_last_stage():
```
gshennvm, line 273: (just a thought): the contract we have with nemo generate is that the outputs it returns will be duplicated across all ranks. 

This breaks in contract -- I think we should move this broadcast PP logic into the strategy itself(since the strategy is already checking if PP is last stage when computing get_length). 

then in the model we can just perform the `assert (response_lengths <= self.cfg.encoder_seq_length).all()` sanity check

nemo_aligner/utils/text_generation_utils.py
odelalleau: In case someone instantiate a `TrackLengthGPTModelTextGenerationStrategy` with an incorrect `context_lengths`. Definitely not supposed to happen with the current code but since this is a standalone class (and I wanted to make sure my code was correct) I preferred to add a check.

nemo_aligner/models/nlp/gpt/megatron_gpt_ppo_actor.py
```diff
@@ -280,19 +262,36 @@ def infer(self, inference_batch):
...
-            tokens=response_tokens,
-            prompt_lengths=prompt_lengths,
-            tokenizer=self.tokenizer,
-            end_strings=self._sampling_params["end_strings"],
-            max_generation_length=self._length_params["max_length"],
-            max_sequence_length=self.cfg.encoder_seq_length,
-        )
+
+        # Sanity check to validate response length.
+        if max_response_length != response_tokens.size(1):
```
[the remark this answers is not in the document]
odelalleau: Hmm not sure how, this check uses the response tokens which the strategy doesn't keep track of. It's really just a sanity check to be sure the response length we compute is consistent with the generation output (that is not fully controlled by the strategy).

nemo_aligner/models/nlp/gpt/megatron_gpt_ppo_actor.py
odelalleau: Good idea, I agree it's nice that way. Done in 02f0a612.

nemo_aligner/models/nlp/gpt/megatron_gpt_ppo_actor.py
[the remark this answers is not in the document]
gshennvm: ah right OK, let's leave it for now

gshennvm (approved):
thanks for the change! this LGTM. This also cleans up our logic on hard coding `<extra_id_1>` which is very welcomed

cc @JRD971000

gshennvm (dismissed):
thanks for the change! this LGTM. This also cleans up our logic on hard coding `<extra_id_1>` which is very welcomed

cc @JRD971000

gshennvm (approved):

PR #44, opened 2023-12-05
add timer impl

gshennvm:
# What does this PR do ?

# Changelog 

# Usage

```python
```

**Pre checks**:

# Additional Information
* Related to # (issue)

### DIFF
--- a/examples/nlp/gpt/train_reward_model.py
+++ b/examples/nlp/gpt/train_reward_model.py
@@ -131,7 +131,7 @@ def main(cfg) -> None:
 
     logger.log_hyperparams(OmegaConf.to_container(cfg))
 
-    timer = Timer(cfg.exp_manager.get("max_time_per_run", None))
+    timer = Timer(cfg.exp_manager.get("max_time_per_run"))
 
     rm_trainer = SupervisedTrainer(
         cfg=cfg.trainer.rm,
--- a/examples/nlp/gpt/train_reward_model.py
+++ b/examples/nlp/gpt/train_reward_model.py
@@ -24,8 +24,8 @@ from nemo_aligner.data.nlp.builders import (
     build_train_valid_test_regression_rm_datasets,
     build_train_valid_test_rm_datasets,
 )
-
 from nemo_aligner.models.nlp.gpt.reward_model_classes import REWARD_MODEL_CLASS_DICT, RewardModelType
+from nemo_aligner.utils.distributed import Timer
 from nemo_aligner.utils.train_script_utils import (
     CustomLoggerWrapper,
     add_custom_checkpoint_callback,
@@ -131,6 +131,8 @@ def main(cfg) -> None:
 
     logger.log_hyperparams(OmegaConf.to_container(cfg))
 
+    timer = Timer(cfg.exp_manager.get("max_time_per_run", None))
+
     rm_trainer = SupervisedTrainer(
         cfg=cfg.trainer.rm,
         model=ptl_model,
@@ -141,6 +143,7 @@ def main(cfg) -> None:
         test_dataloader=None,
         logger=logger,
         ckpt_callback=ckpt_callback,
+        run_timer=timer,
     )
 
     if custom_trainer_state_dict is not None:
--- a/nemo_aligner/algorithms/supervised.py
+++ b/nemo_aligner/algorithms/supervised.py
@@ -18,6 +18,8 @@ from statistics import mean
 import torch
 from omegaconf.dictconfig import DictConfig
 from tqdm import tqdm
+from nemo.utils import logging
+
 
 from nemo_aligner.utils.distributed import SyncTimer
 from nemo_aligner.utils.train_utils import clip_gradients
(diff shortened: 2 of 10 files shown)


examples/nlp/gpt/train_reward_model.py
```diff
@@ -131,6 +131,8 @@ def main(cfg) -> None:
 
     logger.log_hyperparams(OmegaConf.to_container(cfg))
 
+    timer = Timer(cfg.exp_manager.get("max_time_per_run", None))
```
odelalleau, line 134: nit but just a reminder that you can simply write
```suggestion
    timer = Timer(cfg.exp_manager.get("max_time_per_run"))
```

nemo_aligner/utils/distributed.py
```diff
@@ -320,3 +324,37 @@ def sync_and_consume_over_stored_time(self, name=""):
         yield from output_list
 
         del self.stored_results[name]
+
+
+@dataclass
+class Timer:
+    """Timer to tell us when the time limit has reached
```
odelalleau, line 331:
```suggestion
    """Timer to tell us when the time limit is reached
```

nemo_aligner/utils/distributed.py
```diff
@@ -320,3 +324,37 @@ def sync_and_consume_over_stored_time(self, name=""):
...
+    """
+
+    duration: Optional[str]
+
+    def __post_init__(self):
```
odelalleau, line 336: Just curious why make it a dataclass with `__post_init__()` instead of a regular class with a regular `__init__()`?

nemo_aligner/utils/distributed.py
```diff
@@ -320,3 +324,37 @@ def sync_and_consume_over_stored_time(self, name=""):
...
+        self._duration = float("inf")
+
+        if isinstance(self.duration, str):
```
odelalleau, line 339: I think this would be safer:
```suggestion
        if self.duration is not None:
```

nemo_aligner/utils/distributed.py
```diff
@@ -320,3 +324,37 @@ def sync_and_consume_over_stored_time(self, name=""):
...
+            days, hours, mins, seconds = map(int, self.duration.strip().split(":"))
+            self._duration = timedelta(days=days, hours=hours, minutes=mins, seconds=seconds).total_seconds()
+
+    def start_time(self):
+        self._start_time = time.monotonic()
```
odelalleau, line 344: Interesting, I had never see that one! :)

nemo_aligner/algorithms/supervised.py
```diff
@@ -189,6 +197,9 @@ def fit(self):
                     metrics = {k: torch.as_tensor(v) for k, v in metrics.items()}
                     self.save(metrics, is_train_end=is_train_end)
 
+                if run_time_exceeded:
+                    return
```
odelalleau, line 201: 1. Can you add a log that says why we stop
2. Should we call `self.logger.finalize()` before returning?

odelalleau (commented):
Looking good, just a few small questions / comments

nemo_aligner/utils/distributed.py
gshennvm: No reason other than being too lazy to write the `__repr__` etc for the class

nemo_aligner/utils/distributed.py
gshennvm: are you thinking that checking this way let's the users crash immediately if they didn't give a string? whereas if it's another dtype it'll crash during the training run

nemo_aligner/utils/distributed.py
gshennvm: :)

nemo_aligner/algorithms/supervised.py
gshennvm: 1. 07e3990e5ed9ec54d37d3c92e2afda9ee88fd3f4
2. I don't think we need to, the only time we finalize is when the run is successful(which I think here maybe we should consider it as not because we didn't finish training)

nemo_aligner/utils/distributed.py
odelalleau: I don't think it'll crash if for instance someone sets `max_time_per_run` to an integer value, it seems to me it'll just get ignored, right? (though it might crash in `exp_manager()`, not sure)

gshennvm: not sure about exp_manager but yeah it won't crash. Are you saying this is safer because we are implicitly forcing the type to be string here?

nemo_aligner/algorithms/supervised.py
odelalleau: Ok sounds good!

nemo_aligner/utils/distributed.py
odelalleau: Yes -- you can assert it but my main concern was someone accidentally providing a number instead of a string and having it silently ignored (sorry I should have been clearer)

gshennvm:
will carry this change over to DPO, PPO and then undraft. thanks for taking an initial look!

gshennvm:
undrafted! i'm testing it as we speak but let me know what you think

odelalleau (approved):
lgtm, just a minor suggestion

odelalleau (approved):
Good to merge!

PR #45, opened 2023-12-05
use default codeql

gshennvm:
# What does this PR do ?

instead of using a custom codeql configuration, we will use the default one instead. So we don't need this yaml file


odelalleau (approved):
CodeQL is green so I guess it means this works! :D

PR #47, opened 2023-12-05
Public Dockerfile

trias702:
# What does this PR do ?

Adds a public facing dockerfile

# Changelog 

# Usage

```bash
docker image build --network=host -t nemo-aligner:nemo-1.21.0-mlm-0830-TE-1.1 .
```

**Pre checks**:

# Additional Information
* Related to # (issue)


Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+RUN pip install --upgrade-strategy only-if-needed jsonlines
+
+# install TransformerEngine
+RUN pip uninstall -y transformer-engine
+RUN pip install --upgrade git+https://github.com/NVIDIA/TransformerEngine.git@release_v1.1
+# alternatively, if you have issues with TE 1.1, try stable instead
+#RUN pip install --upgrade git+https://github.com/NVIDIA/TransformerEngine.git@stable
+
+# install latest apex
+RUN pip uninstall -y apex
```
yidong72, line 35: can we first install apex then other libraries. The apex takes most of the time to build the container. If we update NeMo or Megatron_LM versions in the dockerfile, it will rebuild the apex.

trias702:
I have updated the Docker file to use the latest NeMo 1.22.0 paired with the latest main branch of Megatron-LM and TransformerEngine 1.1. Confirmed working for DPO training and generation.

Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+
+# MLM
+RUN pip uninstall -y megatron-core
+RUN git clone -b core_r0.4.0 https://github.com/NVIDIA/Megatron-LM.git
+RUN cd Megatron-LM && pip install -e .
+
+WORKDIR /opt
+
+# install the latest NeMo-Aligner
+RUN pip install --upgrade-strategy only-if-needed git+https://github.com/NVIDIA/NeMo-Aligner.git@main
```
yidong72, line 43: I tried this docker file and I got megatron-core version error.
The problem is caused by the fact it installs an old version of megatron-core when running `pip install git+https://github.com/NVIDIA/NeMo-Aligner.git@main`. For some reason we put `nemo_toolkit[nlp]` in the requirement.txt in the aligner which will install an older version of megaton core.
The world around I had is to first install nemo-aligner than install the correct version of megatron-core.
```
# install the latest NeMo-Aligner
RUN pip install --upgrade-strategy only-if-needed git+https://github.com/NVIDIA/NeMo-Aligner.git@main

# MLM
RUN pip uninstall -y megatron-core
RUN git clone -b core_r0.4.0 https://github.com/NVIDIA/Megatron-LM.git
RUN cd Megatron-LM && pip install -e .
```

gshennvm: how about pip install --no-deps? and just install the rest ourselves. We just need nemo and pytrition

yidong72: yes `--no-deps` works too. I tried it as the following
```
# MLM
RUN pip uninstall -y megatron-core
RUN git clone -b core_r0.4.0 https://github.com/NVIDIA/Megatron-LM.git
RUN cd Megatron-LM && pip install -e .

WORKDIR /opt

# install the latest NeMo-Aligner
RUN pip install  --no-deps --upgrade-strategy only-if-needed git+https://github.com/NVIDIA/NeMo-Aligner.git@main
```

trias702:
If you use --no-deps, then it should technically just be: `RUN pip install
--no-deps ***@***.***`

I will commit these changes later today

Dockerfile
trias702: I have updated it to use --no-deps

trias702:
> this dockerfile works for me. are we ready to merge it in?

I would be happy to merge it, but we need @gshennvm to confirm as well

Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+
+# NeMo
+RUN git clone -b r1.22.0 https://github.com/NVIDIA/NeMo.git
+RUN cd NeMo && git cherry-pick -X theirs fa8d416793d850f4ce56bea65e1fe28cc0d092c0
+WORKDIR /opt/NeMo
+# fixes a bug in NeMo when used with latest apex
+RUN sed -i 's/_fast_layer_norm(x, self.weight + 1, self.bias, self.epsilon)/_fast_layer_norm(x, self.weight + 1, self.bias, self.epsilon, False)/g' nemo/collections/nlp/modules/common/megatron/layer_norm_1p.py
+RUN sed -i 's/shutil.rmtree(ckpt_to_dir(filepath))/shutil.rmtree(ckpt_to_dir(filepath), ignore_errors=True)/g' nemo/collections/nlp/parts/nlp_overrides.py
+RUN sed -i 's/\[all\]/\[nlp\]/g' reinstall.sh
+RUN rm -rf .git && ./reinstall.sh
```
gshennvm, line 33: is it possible to upstream this onto nemo? if they ever change this file the sed may start to not work

Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+
+# MLM
+RUN pip uninstall -y megatron-core
+RUN git clone -b core_r0.4.0 https://github.com/NVIDIA/Megatron-LM.git
```
gshennvm, line 37: can we use dockerfile args to specify which MCORE/NeMo/Apex commit we need? this way we can change this dockerfile if so decide, and it's more transparent to the user. something like

```bash
ARG MCORE_COMMIT=some default value
RUN git clone https://github.com/NVIDIA/Megatron-LM.git && \
    cd Megatron-LM && \
    git fetch && \
    if [ ! -z $MCORE_COMMIT ]; then \
        git fetch origin $MCORE_COMMIT && \
        git checkout FETCH_HEAD; \
    fi && \
    pip install -e .
```

Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+RUN cd Megatron-LM && pip install -e .
+
+WORKDIR /opt
+
+# install the latest NeMo-Aligner
+RUN pip install --no-deps git+https://github.com/NVIDIA/NeMo-Aligner.git@main
+
+WORKDIR /workspace
+
+ENV NVTE_APPLY_QK_LAYER_SCALING=1
```
gshennvm, line 47: i don't think we need this in the newer versions

gshennvm (changes requested):
looking good to me, added a few comments.

also can you do these 2 things?
1. add something onto the README that points to this file and discusses how this is a more "stable" dockerfile meant for reproducibility etc, but may not be the latest and greatest
2. update the CHANGELOG so users know we pushed a dockerfile

Dockerfile
trias702: Sadly not for the `_fast_layer_norm` part because Nemo is still using an old version of Apex whereas our Dockerfile uses the latest Apex, hence we need to make that change or else there's an error. Whenever Nemo chooses to upgrade to the latest Apex, they will make that change.

The `all` -> `nlp` we need to change locally as we are reinstalling just the NLP parts, we don't need the ASR and TTS packages.

The shutil.rmtree fix they should add, I'll add a PR for it.

Dockerfile
trias702: Newer versions of which packages? When I test this Dockerfile as is, we still need that ENV declaration

Dockerfile
trias702: Also, these sed commands are quite specifically focused, no wildcards or anything, so if the upstream codes changes at all, these won't find what they're specifically looking for and shouldn't impact anything. I made them to be quite benign in that way.

Dockerfile
trias702: But the way this Dockerfile works is based off of branches, not individual commits. I'm not a huge fan of using individual commit ids as then you get really deep into the weeds and it becomes spaghetti code quickly (which commit ids work with which other ones). I think we really need to ensure our base trinity of Nemo/MLM/TE works with specific tags/branches, and if they don't we really need to fix the upstream code to ensure it does work.

Dockerfile
gshennvm: ah i see, does this change only affect the non mcore path? if that is the case I am okay with even just removing these lines. Since aligner technically doesn't support non mcore(though in theory it runs)

Dockerfile
gshennvm: actually that is totally okay. you can make these tags/branches and it works the same way. Will that be okay?

gshennvm: `MCORE_COMMIT=<<<some tag or branch>>>` works. But i guess then we should do a rename `MCORE_COMMIT -> MCORE_TAG` or something

Dockerfile
gshennvm: i think we need the nemo commit:

https://github.com/NVIDIA/NeMo/commit/a7f0bc1903493888c31436efc2452ff721fa5a67 and it should work

Dockerfile
trias702: The `_fast_layer_norm` is only for non-mcore, but the others are necessary for mcore too.

Dockerfile
trias702: Should I cherry pick that commit into the Dockerfile?

gshennvm: will you be able to use nemo main?  because #71 will break the container if we want to use Aligner main

Dockerfile
gshennvm: ah i see!

so how about this:

line 30: we can remove, because we don't support non-mcore models(and this enforces it)
line 31: this needs to be changed on the nemo side, i've seen customers complain about this
line 32: let's keep it

how does that sound?

trias702: I have removed line 30. Line 31 we need to keep until https://github.com/NVIDIA/NeMo/pull/8130/files gets merged in and I can cherry pick it in. Line 32 left in as requested.

Dockerfile
trias702: Sorted.

Dockerfile
trias702: Sorted now

Dockerfile
```diff
@@ -1,40 +1,71 @@
...
+#RUN cd apex && pip install -v --disable-pip-version-check --no-cache-dir --no-build-isolation --config-settings "--build-option=--cpp_ext" --config-settings "--build-option=--cuda_ext" --config-settings "--build-option=--distributed_adam" --config-settings "--build-option=--permutation_search" --config-settings "--build-option=--bnp" --config-settings "--build-option=--xentropy" --config-settings "--build-option=--focal_loss" --config-settings "--build-option=--index_mul_2d" --config-settings "--build-option=--deprecated_fused_adam" --config-settings "--build-option=--fast_layer_norm" --config-settings "--build-option=--fmha" --config-settings "--build-option=--fast_multihead_attn" --config-settings "--build-option=--transducer" --config-settings "--build-option=--cudnn_gbn" --config-settings "--build-option=--fused_conv_bias_relu" .
```
gshennvm, line 30:
```suggestion
```

Dockerfile
```diff
@@ -1,40 +1,71 @@
...
+	cd apex && \
+    if [ ! -z $APEX_TAG ]; then \
+        git fetch origin $APEX_TAG && \
+        git checkout FETCH_HEAD; \
+    fi && \
+    pip install install -v --no-build-isolation --disable-pip-version-check --no-cache-dir --config-settings "--build-option=--cpp_ext --cuda_ext --fast_layer_norm --distributed_adam --deprecated_fused_adam" ./
 
 # place any util pkgs here
 RUN pip install --upgrade-strategy only-if-needed nvidia-pytriton==0.4.1
 RUN pip install -U --no-deps protobuf==4.24.4
```
gshennvm, line 42: could you also tag these 2 as well?

gshennvm:
looking good! after 2 more minor changes i think we can merge

Dockerfile
trias702: Done

gshennvm (approved):

Issue #53, opened 2023-12-08
Support overwriting model config nodes instead of merging them


odelalleau:
**Is your feature request related to a problem? Please describe.**

This is about supporting something akin to the suggestion in https://github.com/NVIDIA/NeMo-Aligner/pull/48#discussion_r1418209055

PR #55, opened 2023-12-11
overwrite base config attribute

gleibovich-nvidia:
# What does this PR do ?

Creates a mechanism for resolving collisions between attributes found both in checkpoint and training configurations. Addresses the issue with the mechanism suggested in https://github.com/NVIDIA/NeMo-Aligner/issues/53.

# Changelog 

# Usage

```python
```

**Pre checks**:

# Additional Information
* Related to # (issue)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) 
 ### Breaking changes
 
 ### Bug Fixes
-- Fixed a potential issue when the base model's `model.data.data_prefix` config is a list and is about to overriden with
+- Fixed a potential issue when the base model's `model.data.data_prefix` config is a list and is about to be overridden with
 a dictionary from the training configuration.
 
 ## [0.1.0] - 2023-12-04
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) 
 ### Bug Fixes
 - Fixed a potential issue when the base model's `model.data.data_prefix` config is a list and is about to be overridden with
 a dictionary from the training configuration.
+- `exp_manager.max_time_per_run` is now respected, the trainers will save and run validation before exiting if we've reached the time limit.
 
 ## [0.1.0] - 2023-12-04
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) 
 ### Breaking changes
 
 ### Bug Fixes
+- Fixed a potential issue when the base model's `model.data.data_prefix` config is a list and is about to overriden with
+a dictionary from the training configuration.
 
 ## [0.1.0] - 2023-12-04
 ### Added
--- a/examples/nlp/gpt/conf/gpt_dpo.yaml
+++ b/examples/nlp/gpt/conf/gpt_dpo.yaml
@@ -97,6 +97,7 @@ model:
     index_mapping_dir: null # path to save index mapping .npy files, by default will save in the same location as data_prefix
     data_prefix: null
 
+  # define fields from the base model's config that should be ignored when merging with this config.
   overwrite_base_config:
     data:
       data_prefix: True
--- a/examples/nlp/gpt/conf/gpt_dpo.yaml
+++ b/examples/nlp/gpt/conf/gpt_dpo.yaml
@@ -97,4 +97,8 @@ model:
     index_mapping_dir: null # path to save index mapping .npy files, by default will save in the same location as data_prefix
     data_prefix: null
 
+  overwrite_base_config:
+    data:
+      data_prefix: True
+
 precision: ${trainer.precision}
--- a/examples/nlp/gpt/conf/gpt_ppo_critic.yaml
+++ b/examples/nlp/gpt/conf/gpt_ppo_critic.yaml
@@ -96,8 +96,4 @@ model:
       constant_steps: 1000
       min_lr: 9e-7
 
-  overwrite_base_config:
-    data:
-      data_prefix: True
-
   precision: ${trainer.precision}
--- a/examples/nlp/gpt/conf/gpt_ppo_critic.yaml
+++ b/examples/nlp/gpt/conf/gpt_ppo_critic.yaml
@@ -96,4 +96,8 @@ model:
       constant_steps: 1000
       min_lr: 9e-7
 
+  overwrite_base_config:
+    data:
+      data_prefix: True
+
   precision: ${trainer.precision}
--- a/examples/nlp/gpt/conf/gpt_sft.yaml
+++ b/examples/nlp/gpt/conf/gpt_sft.yaml
@@ -165,6 +165,10 @@ model:
       constant_steps: 1000
       min_lr: 9e-7
 
+  overwrite_base_config:
+    data:
+      data_prefix: True
+
 inference:
   greedy: True # Whether or not to use sampling ; use greedy decoding otherwise
   top_k: 0  # The number of highest probability vocabulary tokens to keep for top-k-filtering.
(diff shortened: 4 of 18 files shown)


nemo_aligner/utils/utils.py
```diff
@@ -123,7 +123,34 @@ def load_and_override_model_config(restore_path, model_cfg_to_overwrite, remove_
...
+    def recursive_merge(base, overwrite, path=[]):
+        for key, value in overwrite.items():
+            current_path = path + [key]
+
+            if key not in base:
+                continue
+
+            if isinstance(value, DictConfig) and isinstance(base[key], DictConfig):
+                recursive_merge(base[key], value, path=current_path)
+            else:
```
odelalleau, line 141: We should check if `value` is True (one reason to make it a config is so we can easily set it to False if needed)

nemo_aligner/utils/utils.py
```diff
@@ -123,7 +123,34 @@ def load_and_override_model_config(restore_path, model_cfg_to_overwrite, remove_
...
+                current_config = overwrite_base_config
+                for sub_key in current_path:
+                    if isinstance(current_config, DictConfig):
+                        current_config = current_config[sub_key] if sub_key in current_config else None
```
odelalleau, line 145: I'm a bit confused by what this `current_config` is doing. Since we're going down the same path we went through aren't we guaranteed that `overwrite_base_config` has such a path? And isn't `current_config` essentially the same as `value`?

nemo_aligner/utils/utils.py
```diff
@@ -123,7 +123,34 @@ def load_and_override_model_config(restore_path, model_cfg_to_overwrite, remove_
...
         checkpoint_cfg.pop("nemo_version", None)
 
-    return OmegaConf.merge(checkpoint_cfg, model_cfg_to_overwrite)
+    resolved_checkpoint_cfg = resolve_overlapping_fields(checkpoint_cfg, model_cfg_to_overwrite)
+
+    return OmegaConf.merge(resolved_checkpoint_cfg, model_cfg_to_overwrite)
+
+
+def resolve_overlapping_fields(base_config, overwrite_config):
+    def recursive_merge(base, overwrite, path=[]):
```
odelalleau, line 132: Just to be safe, better avoid a mutable default (using `None` instead)

nemo_aligner/utils/utils.py
```diff
@@ -123,7 +123,34 @@ def load_and_override_model_config(restore_path, model_cfg_to_overwrite, remove_
...
+                    if isinstance(current_config, DictConfig):
+                        current_config = current_config[sub_key] if sub_key in current_config else None
+
+                if current_config is not None:
+                    base.pop(key, None)
+
+    overwrite_base_config = overwrite_config.get("overwrite_base_config", OmegaConf.create())
+    recursive_merge(base_config, overwrite_config)
+
+    return DictConfig(base_config)
```
odelalleau, line 153: Why call `DictConfig` here? We've already mutated the input base config (may be safer not to => deep copy it first instead?)

odelalleau (commented):
The high-level logic seems fine but I find the main function doing the resolution a bit hard to understand. I feel like it can probably be made simpler.

gleibovich-nvidia:
> The high-level logic seems fine but I find the main function doing the resolution a bit hard to understand. I feel like it can probably be made simpler.

Yeah I agree it's confusing. I think it's mainly because of limited APIs from `OmegaConf.DictConfig`. Does addressing the comments raised made it clearer now? Did you had some other logic in mind which would have made it simpler?

nemo_aligner/utils/utils.py
odelalleau: Resolving this comment as I had misunderstood this code.

nemo_aligner/utils/utils.py
odelalleau: Resolving as this is addressed in my suggested commit

odelalleau (commented):
> Does addressing the comments raised made it clearer now? Did you had some other logic in mind which would have made it simpler?

It was a bit difficult for me to wrap my head around this yesterday without actually running the code. After running it I came up with 9487d754 which you can cherry-pick on top of your branch if you like it.

This commit:
* Makes things simpler overall (IMO) -- no more nested recursive function dealing with 3 configs, we just walk down `overwrite_base_config` along with the base config to identify which keys should be removed
* Fixes a couple of issues: (1) the value of the flag (True or False) used to be ignored and (2) the keys being popped were the children of the key we actually want to pop, which I believe wouldn't have worked with a `ListConfig`
* Actually modifies `base_config` in place: I know I suggested not to do this in my previous review, but I hadn't realized that this base config  was created from within `load_and_override_model_config()` and thus it seems safe to me (while keeping the code simpler)
* Removes the `overwrite_base_config` key from the final model config (this is just to avoid cluttering the config -- we don't need this key anymore once the config is merged)

Let me know what you think!

examples/nlp/gpt/conf/gpt_dpo.yaml
```diff
@@ -97,4 +97,8 @@ model:
     index_mapping_dir: null # path to save index mapping .npy files, by default will save in the same location as data_prefix
     data_prefix: null
 
+  overwrite_base_config:
```
odelalleau, line 100: Let's add a comment to make it more understandable (please do it in all files -- feel free to rewrite):

```suggestion
  # Which fields from the base model config should be ignored when merging with this config.
  overwrite_base_config:
```

examples/nlp/gpt/conf/inference_rm.yaml
```diff
@@ -24,4 +24,8 @@ model:
   
   regression:
     merge_attributes: False # whether to merge attribute values into a scalar
-    attribute_weights: null # apply these weights to each attributes when merging them into a scalar
\ No newline at end of file
+    attribute_weights: null # apply these weights to each attributes when merging them into a scalar
+
+  overwrite_base_config:
```
odelalleau, line 29: Can be removed from this file (it doesn't have a `data_prefix`)

examples/nlp/gpt/conf/gpt_ppo_critic.yaml
```diff
@@ -96,4 +96,8 @@ model:
       constant_steps: 1000
       min_lr: 9e-7
 
+  overwrite_base_config:
```
odelalleau, line 99: Can be removed from this file (it doesn't have a `data_prefix`)

examples/nlp/gpt/conf/gpt_sft.yaml
```diff
@@ -165,6 +165,10 @@ model:
       constant_steps: 1000
       min_lr: 9e-7
 
+  overwrite_base_config:
```
odelalleau, line 168: Can be removed from this file (it doesn't have a `data_prefix`)

odelalleau (changes requested):
Looks good, just some final suggestions before merging.

In addition to the comments below, please also add a CHANGELOG entry (it's a bugfix, something like "fix crash that could occur when the base model's `data.data_prefix` config was a list")

gleibovich-nvidia:
> Looks good, just some final suggestions before merging.
> 
> In addition to the comments below, please also add a CHANGELOG entry (it's a bugfix, something like "fix crash that could occur when the base model's `data.data_prefix` config was a list")

Sure, done.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Breaking changes
 
 ### Bug Fixes
+- Fixed a potential issue when the base model's `model.data.data_prefix` config is a list and is about to overriden with
```
odelalleau, line 13:
```suggestion
- Fixed a potential issue when the base model's `model.data.data_prefix` config is a list and is about to be overridden with
```

odelalleau (dismissed):
Please just fix the minor typo in changelog then good to merge!

odelalleau (dismissed):
Sorry I didn't realize I needed to re-review, all good and thanks a lot for the help!

gleibovich-nvidia:
> Sorry I didn't realize I needed to re-review, all good and thanks a lot for the help!

Merged `main` into the PR, and now need to re-review (again) in order to merge :-)

odelalleau (dismissed):
haha :D

odelalleau (approved):
Fun times!

PR #78, opened 2024-01-11
Add Conservative DPO, IPO, and KTO

ertkonuk:
# What does this PR do ?

Adds Conservative DPO (CDPO), IPO, and KTO methods

# Changelog 

# Usage
* For CDPO, simply set the model.dpo.label_smoothing to a positive non-zero value. For IPO, and KTO, set the model.dpo.preference_loss to "ipo" or "kto", respectively.

```python
```

**Pre checks**:
- [x] Make sure you read and followed [Contributor guidelines](/CONTRIBUTING.md)

- [x] Does the trainer resume and restore model state all states?
- [x] Does the trainer support all parallelism techniques(PP, TP, DP)?
- [x] Does the trainer support `max_steps=-1` and `validation`?
- [x] Does the trainer only call APIs defined in [alignable_interface.py](/nemo_aligner/models/alignable_interface.py)?
- [x] Does the trainer have proper logging?

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
+++ b/nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
@@ -62,7 +62,7 @@ class MegatronGPTDPOModel(MegatronGPTModel, SupervisedInterface):
         self.label_smoothing = self.cfg.dpo.get("label_smoothing", 0.0)
         self.preference_loss = self.cfg.dpo.get("preference_loss", "dpo")
 
-        supported_loss_functions = ['dpo','ipo']
+        supported_loss_functions = ["dpo","ipo","kto"]
         if self.preference_loss not in supported_loss_functions:
             raise ValueError(
                 f"Preference loss {self.preference_loss} is not supported. Supported loss functions are {supported_loss_functions}"
@@ -204,14 +204,27 @@ class MegatronGPTDPOModel(MegatronGPTModel, SupervisedInterface):
 
         chosen_rewards, reject_rewards = self.split_output_tensor(rewards)
         logits = chosen_rewards - reject_rewards
-        assert self.preference_loss in ["dpo", "ipo"]
+        assert self.preference_loss in ["dpo", "ipo", "kto"]
         if self.preference_loss == "dpo":
             loss = (
                 -torch.nn.functional.logsigmoid(self.ref_policy_kl_penalty * logits) * (1.0 - self.label_smoothing)
                 -torch.nn.functional.logsigmoid(-self.ref_policy_kl_penalty * logits) * self.label_smoothing
             )
         elif self.preference_loss == "ipo":
+            # self.ref_policy_kl_penalty is the tau parameter in the paper
             loss = (logits - 1.0 / (2.0 * self.ref_policy_kl_penalty)) ** 2
+        elif self.preference_loss == "kto":
+            rewards_kl = self.get_reduced_masked_logps(
+                 pi_logprobs - ref_logprobs, labels, average_log_probs=True
+            )            
+            chosen_kl, reject_kl = self.split_output_tensor(rewards_kl)
+            loss = torch.cat(
+                (
+                    1.0 - torch.nn.functional.sigmoid(self.ref_policy_kl_penalty * (chosen_rewards - reject_kl.clamp(min=0))),
+                    1.0 - torch.nn.functional.sigmoid(self.ref_policy_kl_penalty * (chosen_kl.clamp(min=0) - reject_rewards)),
+                ),
+                0,
+            ).mean()
         else:
             raise ValueError(f"The loss function {self.preference_loss} is not supported.")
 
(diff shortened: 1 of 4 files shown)


nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
```diff
@@ -193,9 +201,36 @@ def loss_func(self, pi_logprobs, ref_logprobs, labels, average_log_probs=False):
...
+                -torch.nn.functional.logsigmoid(self.ref_policy_kl_penalty * logits) * (1.0 - self.label_smoothing)
+                - torch.nn.functional.logsigmoid(-self.ref_policy_kl_penalty * logits) * self.label_smoothing
+            )
+        elif self.preference_loss == "ipo":
+            # self.ref_policy_kl_penalty is the tau parameter in the paper
+            loss = (logits - 1.0 / (2.0 * self.ref_policy_kl_penalty)) ** 2
+        elif self.preference_loss == "kto":
+            rewards_kl = self.get_reduced_masked_logps(pi_logprobs - ref_logprobs, labels, average_log_probs=True)
+            chosen_kl, reject_kl = self.split_output_tensor(rewards_kl)
+            loss = torch.cat(
```
gshennvm, line 219: another nit: I'm not sure what the performance impact of this is but this is creating a tensor and then filling it in with these 2 things followed by a mean -> why not do .sum() on each of these tensors and then divide it ourselves?

I think it also makes it more readable that way

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
```diff
@@ -193,9 +201,36 @@ def loss_func(self, pi_logprobs, ref_logprobs, labels, average_log_probs=False):
         rewards = self.get_reduced_masked_logps(
             pi_logprobs - ref_logprobs, labels, average_log_probs=average_log_probs
         )
-        chosen_rewards, reject_rewards = self.split_output_tensor(self.ref_policy_kl_penalty * rewards)
 
-        loss = -torch.nn.functional.logsigmoid(chosen_rewards - reject_rewards)
+        chosen_rewards, reject_rewards = self.split_output_tensor(rewards)
+        logits = chosen_rewards - reject_rewards
+        assert self.preference_loss in ["dpo", "ipo", "kto"]
```
gshennvm, line 207: can you put the implementation of these 3 loss function in separate functions, and then put these functions in a dictionary?

something like:


```python
PREFERENCE_LOSS_FUNCTIONS = {
"dpo": dpo_loss_function,
...
}
```
and then get it in the model `__init__`?

gshennvm (changes requested):

gshennvm:
with this addition, we should also rename dpo -> something else to make it clear we have the option to do other things.

currently im thinking of calling this gpt_preference_optimization @odelalleau do you have any thoughts on a better name?

gshennvm:
@ertkonuk please also make sure that we reflect this change on our DPO [tutorial](https://github.com/NVIDIA/NeMo-Aligner/blob/main/docs/user-guide/DPO.rst), and I will ping you about adding these loss functions onto our CI once we finalize the review.

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
```diff
@@ -193,9 +201,36 @@ def loss_func(self, pi_logprobs, ref_logprobs, labels, average_log_probs=False):
...
+        if self.preference_loss == "dpo":
+            loss = (
+                -torch.nn.functional.logsigmoid(self.ref_policy_kl_penalty * logits) * (1.0 - self.label_smoothing)
```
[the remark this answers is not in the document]
ertkonuk: We don't need to compute that twice. I changed the code.

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
ertkonuk: Changed it!

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
ertkonuk: I put this dict outside the class. Alternatively, we can also make it a part of the MegatronGPTDPOModel class.

odelalleau:
> with this addition, we should also rename dpo -> something else to make it clear we have the option to do other things.
> 
> currently im thinking of calling this gpt_preference_optimization @odelalleau do you have any thoughts on a better name?

I'm not sure we should rename it. I feel like it may cause more trouble and confusion than keeping the DPO name and documenting it as "DPO and related algorithms / variants" => I wouldn't rush a renaming right now given the limited scope of the changes in this PR (and this is not a criticism: it's great that it's limited!)

Btw apologies but my full review may come late => feel free to merge without me. I want to spend some time actually looking at these papers, but this won't be possible until next week.

At first glance the code structure looks good to me!

kawine:
thanks for including kto in this @ertkonuk ! 

this kto implemention is the kto-paired version in huggingface that assumes access to a paired preference data. the more powerful (and standard) version of kto can work with purely biary data (+1/-1, good/bad) and supports extreme data imbalances (e.g., 5% positive examples and 95% negative examples), and has some minor changes to make training more stable
- huggingface is close to merging the PR for the unpaired version (see [here](https://github.com/huggingface/trl/pull/1181#discussion_r1492169068))
- [open-rlhf](https://github.com/OpenLLMAI/OpenRLHF) has already implemented this version (as has our [own repo](https://github.com/ContextualAI/HALOs))

i think NeMo users would find it more useful to have the latter version of KTO, since it would allow them to align with a much more available kind of feedback

SahilJain314:
It looks like there are large discrepancies between fp32 and bf16 runs at the moment. Looking into it.

Cppowboy:
When will this PR be merged?

gshennvm:
> Cppowboy

we have some(but not all) of these algorithms here: https://github.com/NVIDIA/NeMo-Aligner/pull/195 and currently working on adding KTO + verification

PR #89, opened 2024-01-23
Fix crash when using a separate reward model server in PPO

odelalleau:
The problem was that PyTriton expects all outputs to have the same batch size, while the critic was returning a dummy length-1 tensor for rewards. It is cleaner to just not reeturn rewards at all.

@gshennvm should we add this setup to CI?

### DIFF
--- a/nemo_aligner/algorithms/critic_server_trainer.py
+++ b/nemo_aligner/algorithms/critic_server_trainer.py
@@ -72,11 +72,12 @@ class CriticServerTrainer:
             Tensor(name="sequence_lengths", shape=(-1,), dtype=np.int64, optional=True),
             Tensor(name="add_EOS", shape=(1,), dtype=np.bool_, optional=True),
         )
-        self.infer_outputs = (
+        self.infer_outputs = [
             Tensor(name="values", shape=(-1,), dtype=np.float32),
-            Tensor(name="rewards", shape=(-1,), dtype=np.float32),
             Tensor(name="exceeded", shape=(1,), dtype=np.int32),
-        )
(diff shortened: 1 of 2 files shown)


nemo_aligner/algorithms/critic_server_trainer.py
```diff
@@ -72,11 +72,12 @@ def __init__(self, cfg, model, optimizer, scheduler, logger, ckpt_callback, gbs)
             Tensor(name="sequence_lengths", shape=(-1,), dtype=np.int64, optional=True),
             Tensor(name="add_EOS", shape=(1,), dtype=np.bool_, optional=True),
         )
-        self.infer_outputs = (
+        self.infer_outputs = [
             Tensor(name="values", shape=(-1,), dtype=np.float32),
```
gshennvm, line 76: nit: can we do the same logic but with "values" key instead?

it feels more intuitive to add on the "value" key rather than the "reward" key when  `combine_rm_and_critic_server=True`

odelalleau: Sorry, I'm not sure I follow -- values are always returned by the critic, what do you mean exactly?

gshennvm: ah! my appologies, I don't know what I was thinking -- this looks good to me

gshennvm:
> @gshennvm should we add this setup to CI?

Yes, currently there's some issues with the CI that is blocking CI dev work. I'll make a note to add it

gshennvm (approved):

PR #92, opened 2024-01-25
change dockerfile for ci

gshennvm:
# What does this PR do ?

logic of dockerfile does not change but need to change the ALIGNER_COMMIT arg so it's easier to run on the nemo-ci

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,6 +8,7 @@ ARG MLM_TAG=core_r0.4.0
 ARG NEMO_TAG=r1.22.0
 ARG PYTRITON_VERSION=0.4.1
 ARG PROTOBUF_VERSION=4.24.4
+ARG ALIGNER_COMMIT=main
 
 # if you get errors building TE or Apex, decrease this to 4
 ARG MAX_JOBS=8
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -68,9 +69,13 @@ RUN pip uninstall -y megatron-core && \
     fi && \
     pip install -e .
 
-WORKDIR /opt
-
-# install the latest NeMo-Aligner
-RUN pip install --no-deps git+https://github.com/NVIDIA/NeMo-Aligner.git@main
+RUN git clone https://github.com/NVIDIA/NeMo-Aligner.git && \
```
trias702, line 72: I would add a comment above `# install the latest NeMo-Aligner`

trias702 (commented):
LGTM, but you deleted the comment for the Aligner section and I would recommend adding it back

Dockerfile
gshennvm: ah right, how's this? 4bac22fdfc57082f0f8bb83c6a80cf8f50f78b2c

trias702 (approved):
LGTM

PR #93, opened 2024-01-29
moving _str_to_dtype

gshennvm:
upgrade to a newer nemo version, which moved the function and changed it to a dictionary


nemo_aligner/models/nlp/gpt/megatron_gpt_reward_model.py
```diff
@@ -31,8 +31,8 @@
     get_ltor_masks_and_position_ids,
 )
 from nemo.collections.nlp.parts.utils_funcs import get_last_rank
-from nemo.core.optim.distributed_adam import _str_to_dtype
 from nemo.utils import AppState, logging
+from nemo.utils.dtype import _str_to_dtype
```
odelalleau, line 35: Can you use the new function `str_to_dtype()` instead of the private dict?

PR #97, opened 2024-02-02
Lora for SFT/SteerLM

HeyyyyyyG:
# What does this PR do ?

Add LoRA support for SFT/SteerLM
Only support DP and TP
Have to change Dockerfile to cherry pick a new commit in NeMo to support O2

# Changelog 

# Usage

```python
```

**Pre checks**:

# Additional Information
* Related to # (issue)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -56,6 +56,7 @@ RUN git clone https://github.com/NVIDIA/NeMo.git && \
     git cherry-pick --no-commit -X theirs \
         fa8d416793d850f4ce56bea65e1fe28cc0d092c0 \
         a7f0bc1903493888c31436efc2452ff721fa5a67 \
+        52d50e9e09a3e636d60535fd9882f3b3f32f92ad \
         7d3d9ac3b1aecf5786b5978a0c1e574701473c62 && \
     sed -i 's/shutil.rmtree(ckpt_to_dir(filepath))/shutil.rmtree(ckpt_to_dir(filepath), ignore_errors=True)/g' nemo/collections/nlp/parts/nlp_overrides.py && \
     rm -rf .git && pip install -e ".[nlp]" && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,7 +57,7 @@ RUN git clone https://github.com/NVIDIA/NeMo.git && \
         fa8d416793d850f4ce56bea65e1fe28cc0d092c0 \
         a7f0bc1903493888c31436efc2452ff721fa5a67 \
         52d50e9e09a3e636d60535fd9882f3b3f32f92ad \
-        3524ff4a1efe01e95253318493b166cf7fb27e99 \
+        9940ec60058f644662809a6787ba1b7c464567ad \
         7d3d9ac3b1aecf5786b5978a0c1e574701473c62 && \
     sed -i 's/shutil.rmtree(ckpt_to_dir(filepath))/shutil.rmtree(ckpt_to_dir(filepath), ignore_errors=True)/g' nemo/collections/nlp/parts/nlp_overrides.py && \
     rm -rf .git && pip install -e ".[nlp]" && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,6 +57,7 @@ RUN git clone https://github.com/NVIDIA/NeMo.git && \
         fa8d416793d850f4ce56bea65e1fe28cc0d092c0 \
         a7f0bc1903493888c31436efc2452ff721fa5a67 \
         52d50e9e09a3e636d60535fd9882f3b3f32f92ad \
+        3524ff4a1efe01e95253318493b166cf7fb27e99 \
         7d3d9ac3b1aecf5786b5978a0c1e574701473c62 && \
     sed -i 's/shutil.rmtree(ckpt_to_dir(filepath))/shutil.rmtree(ckpt_to_dir(filepath), ignore_errors=True)/g' nemo/collections/nlp/parts/nlp_overrides.py && \
     rm -rf .git && pip install -e ".[nlp]" && \
--- a/examples/nlp/gpt/train_gpt_sft.py
+++ b/examples/nlp/gpt/train_gpt_sft.py
@@ -128,9 +128,9 @@ def main(cfg) -> None:
 
     assert cfg.model.peft.peft_scheme in ["lora", "none"], "Only support LoRA or Full finetuning"
     if cfg.model.peft.peft_scheme == "lora":
-        assert cfg.model.megatron_amp_O2 == "False", "LoRA doesn't support O2"
+        assert cfg.model.megatron_amp_O2 == False, "LoRA doesn't support O2"
         assert (
-            cfg.optim.name != "distributed_fused_adam"
+            cfg.model.optim.name != "distributed_fused_adam"
         ), "LoRA doesn't support distributed_fused_adam, please use fused_adam"
 
     ptl_model, updated_cfg = load_from_nemo(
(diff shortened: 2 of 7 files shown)


nemo_aligner/utils/utils.py
```diff
@@ -90,7 +97,7 @@ def load_from_nemo(
         save_restore_connector=connector,
         strict=strict,
     )
-    return model
+    return model, model_cfg if return_updated_cfg else model
```
gshennvm, line 100: if you want the model_cfg can you use this function instead?

[load_checkpoint_model_config](https://github.com/NVIDIA/NeMo-Aligner/blob/main/nemo_aligner/utils/utils.py#L96)

nemo_aligner/utils/train_script_utils.py
```diff
@@ -95,6 +95,7 @@ def disable_data_callbacks(ptl_model, train_dataloader, train_ds):
 def init_using_ptl(ptl_trainer, ptl_model, train_dataloader, train_ds):
     """initializes the model using PTL
     """
+    ptl_model.setup_complete = True
```
gshennvm, line 98: nit: can this be moved to the bottom of the init_using_ptl function?

examples/nlp/gpt/train_gpt_sft.py
```diff
@@ -119,15 +126,34 @@ def main(cfg) -> None:
...
+    if cfg.model.peft.restore_from_path is not None:
+        # initialize peft weights from a checkpoint instead of randomly
+        # This is not the same as resume training because optimizer states are not restored.
+        logging.info("PEFT Weights will be loaded from", cfg.model.peft.restore_from_path)
+        ptl_model.load_adapters(cfg.model.peft.restore_from_path, peft_cfg_cls(updated_cfg))
+    elif peft_cfg_cls is not None:
+        logging.info("Adding adapter weights to the model for PEFT")
+        ptl_model.add_adapter(peft_cfg_cls(updated_cfg))
+    else:
+        logging.info(f"Running full finetuning since no peft scheme is given.\n{ptl_model.summarize()}")
```
gshennvm, line 155: just making sure -- when PTL resumes it will automatically figure out how to load the LoRa weights and overwrite whatever we did here right?

gshennvm, line 155: i believe we also want to use PEFT for other algorithms as well, so I think we should create a `initialize_peft` function in `train_script_utils` to handle all this logic related to PEFT(such as error checking, and loading the adaptors)

examples/nlp/gpt/train_gpt_sft.py
```diff
@@ -119,15 +126,34 @@ def main(cfg) -> None:
     with open_dict(cfg):
         cfg.model.precision = cfg.trainer.precision
 
-    ptl_model = load_from_nemo(
+    assert cfg.model.peft.peft_scheme in ["lora", "none"], "Only support LoRA or Full finetuning"
```
gshennvm, line 129: maybe the `initialize_peft` function should take in a peft config

examples/nlp/gpt/train_gpt_sft.py
```diff
@@ -119,15 +126,34 @@ def main(cfg) -> None:
...
+    if cfg.model.peft.peft_scheme == "lora":
+        assert (
+            cfg.model.optim.name != "distributed_fused_adam"
+        ), "LoRA doesn't support distributed_fused_adam, please use fused_adam"
```
gshennvm, line 133: I think we shouldn't have to raise this on aligner in case nemo ever fixes it. @cuichenx will do a check for this on the nemo side. we can pick him his commit once done

gshennvm (changes requested):
thanks for the change! i added some comments, and once we resolve we should add some small tests onto the CI

Dockerfile
```diff
@@ -56,6 +56,7 @@ RUN git clone https://github.com/NVIDIA/NeMo.git && \
     git cherry-pick --no-commit -X theirs \
         fa8d416793d850f4ce56bea65e1fe28cc0d092c0 \
         a7f0bc1903493888c31436efc2452ff721fa5a67 \
+        52d50e9e09a3e636d60535fd9882f3b3f32f92ad \
```
gshennvm, line 59: in addition, picking in this commit will be good https://github.com/NVIDIA/NeMo/pull/8323/files

examples/nlp/gpt/train_gpt_sft.py
HeyyyyyyG: Cool, didn't notice the check in nemo

HeyyyyyyG: 605bad8baf13e6f5107f9e330f79b269fb8866dd

examples/nlp/gpt/train_gpt_sft.py
HeyyyyyyG: 09ad8c61852debbbd2e713396f4b99f1441b52f5

examples/nlp/gpt/train_gpt_sft.py
HeyyyyyyG: Good point, 09ad8c61852debbbd2e713396f4b99f1441b52f5

HeyyyyyyG: yes

nemo_aligner/utils/train_script_utils.py
HeyyyyyyG: No we can't because it will cause an error when resume training. I move it to `init_peft` function, maybe that looks better? 
605bad8baf13e6f5107f9e330f79b269fb8866dd

nemo_aligner/utils/utils.py
HeyyyyyyG: What we need here is the model_cfg after applying `modify_config_fn` , I don't see how we can replace with this function

nemo_aligner/utils/train_script_utils.py
```diff
@@ -135,6 +137,28 @@ def extract_optimizer_scheduler_from_ptl_model(ptl_model):
     return ptl_model.optimizers().optimizer, ptl_model.lr_schedulers()
 
 
+def init_peft(ptl_model, updated_cfg):
```
gshennvm, line 140: nit: can this updated_cfg be called `peft_cfg`?

examples/nlp/gpt/train_gpt_sft.py
```diff
@@ -119,15 +126,18 @@ def main(cfg) -> None:
...
         GPTSFTModel,
         cfg,
         trainer,
         strict=True,
         modify_config_fn=_modify_config,
         restore_path=cfg.model.restore_from_path,
+        return_updated_cfg=True,
     )
 
+    init_peft(ptl_model, updated_cfg)
```
gshennvm, line 139: we should just give the peft cfg

nemo_aligner/utils/utils.py
gshennvm: ah sorry good point -- the SFT configs work a bit differently than the other trainers(a long standing TODO to fix). this shouldn't block your PR so resolving it

examples/nlp/gpt/train_gpt_sft.py
HeyyyyyyG: Unfortunately this is not only peft cfg, it also contains base model cfg, because we need those (e.g num_attention_heads, hidden_size, num_layers) to initialize lora modules in [nemo](https://github.com/NVIDIA/NeMo/blob/5fdd12e9a9711b241023eb4ed0922733d69ded5e/nemo/collections/nlp/parts/peft_config.py#L69).

gshennvm: ah okay. then i'm good with this.

gshennvm:
BEFORE MERGE: I am triggering a pipeline to check SFT with your commit. I'll approve once it passes.

gshennvm (approved):
tests passed thanks!

PR #108, opened 2024-02-16
SPIN Algorithm

trias702:
# What does this PR do ?

Implements the SPIN algorithm into Nemo Aligner

# Changelog 
- Please update the [CHANGELOG.md](/CHANGELOG.md) under next version with high level changes in this PR. (done)

# Usage

```python
```

**Pre checks**:

- [X] Does the trainer resume and restore model state all states?
- [X] Does the trainer support all parallelism techniques(PP, TP, DP)?
- [X] Does the trainer support `max_steps=-1` and `validation`?
- [X] Does the trainer only call APIs defined in [alignable_interface.py](/nemo_aligner/models/alignable_interface.py)?
- [X] Does the trainer have proper logging?

# Additional Information
* Related to # (issue)


nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
```diff
@@ -0,0 +1,524 @@
...
+        }
+
+        # move to CPU
+        metrics = {k: v.item() for k, v in metrics.items()}
+
+        return loss_mean.item(), metrics
+
+    def get_loss_and_metrics_vanilla_sft(self, batch, forward_only):
+        """Take a data_iter which is an interator over the microbatches
+            and return loss as well as metrics
```
odelalleau, line 293: This takes a `batch` as input, not a `data_iter`

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
```diff
@@ -0,0 +1,524 @@
...
+        fwd_bwd_function = get_forward_backward_func()
+        fwd_loss_fn = super().get_forward_output_and_loss_func(forward_only)
+
+        losses_reduced = fwd_bwd_function(
+            forward_step_func=fwd_loss_fn,
+            data_iterator=data_iter,
+            model=self.model,
+            num_microbatches=get_num_microbatches(),
+            forward_only=forward_only,
+            micro_batch_size=self.cfg.micro_batch_size,
```
odelalleau, line 314: Shouldn't the micro-batch size be taken from `validation_ds.micro_batch_size`? (unless we get rid of that one)

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
```diff
@@ -0,0 +1,524 @@
...
+                    if ref_policy is not None:
+                        self.ref_policy_state_dict = ref_policy
+                    module.load_state_dict(checkpoint_state_dict, strict=True)
+            else:
+                # when restoring a distributed checkpoint from a ptl checkpoint we need to defer loading the state_dict
+                # see NLPModel.on_load_checkpoint
+                checkpoint["state_dict"] = {}
+
+        # legacy checkpointing for interleaved
+        else:
```
odelalleau, line 444: Can we just drop support for legacy checkpoints? We don't want to have to implement and test this scenario.

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
```diff
@@ -0,0 +1,524 @@
...
+
+        # add in the reference policy weights
+        if self.ref_policy_state_dict is not None:
+            sharded_state_dict["reference_policy"] = make_sharded_tensors_for_checkpoint(
+                self.ref_policy_state_dict, prefix
+            )
+
+        return sharded_state_dict
+
+    def on_load_checkpoint(self, checkpoint) -> None:
```
odelalleau, line 415: I guess there are still some issues with re-loading the reference policy's state dict based on what you said earlier this week. At first sight I'm not sure where the issue is, but I'm not familiar enough with this distributed checkpointing stuff.

One thing I was wondering is you may remember our discussion where I mentioned https://github.com/NVIDIA/NeMo/blob/bb575b72fd0be51ae10cc77d9f89ddb9e9d3b96d/nemo/collections/nlp/parts/nlp_overrides.py#L338 which erases `state_dict`. So I'm curious why you would still use it here... (but again, I really don't have a full understanding of the code flow).

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
```diff
@@ -0,0 +1,524 @@
...
+    prepare_for_training_step,
+    prepare_for_validation_step,
+    set_eval,
+    set_sync_funcs,
+    set_train,
+)
+from nemo_aligner.utils.utils import configure_batch_sizes, cpu_weight_swap, offload_distributed_adam
+
+
+class MegatronGPTSPINModel(MegatronGPTModel, SupervisedInterface):
```
odelalleau, line 48: Looks like SPIN is essentially a mix of DPO + PPO ==> this creates a lot of copy/paste.

It'd be nice to try and refactor the code to promote more code re-use. Would be non trivial though, and we may want to postpone that until after the class design refactor that @gshennvm has been looking into.

odelalleau (changes requested):
Looks good to me overall, mostly a bunch of small polish tweaks. There's a higher-level design comment to think about regarding how to better re-use existing code instead of copy/pasting from DPO / SFT / PPO, but as I mentioned I'm ok to have it addressed later (hopefully...)

nemo_aligner/algorithms/spin.py
```diff
@@ -0,0 +1,504 @@
...
+                self.model, megatron_amp_O2=self.model.cfg.get("megatron_amp_O2", False)
+            )
+
+        self.logger.finalize()
+
+    def save(self, extra_candidates=None, is_train_end=False):
+        self.model.prepare_for_training()
+        # load back in the adam states if needed
+        torch.cuda.synchronize()
+        torch.distributed.barrier()
```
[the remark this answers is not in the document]
gshennvm: prepare for training loads back the training states from CPU -> GPU(i.e the adam states) so PTL save can save the adam optimizer. I think synchronize is probably not needed but added just to be sure all the hosttodevices are done.

In theory though, we don't need to call this to get the adam states back since they are all on CPU, but it crashes if it's not on GPU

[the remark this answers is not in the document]
odelalleau: I see, thanks. I guess just moving the comment one line above would already help make it clearer!

examples/nlp/gpt/conf/gpt_spin.yaml
```diff
@@ -0,0 +1,182 @@
...
+
+trainer:
+  num_nodes: 8
+  devices: 8
+  accelerator: gpu
+  precision: bf16-mixed
+  
+  # spin specific args
+  spin:
+    max_iterations: 3
```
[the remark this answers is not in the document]
trias702: max_iterations is T from the paper. It represents how many rounds the game is played for, and each round can be played for multiple epochs

nemo_aligner/data/nlp/builders.py
```diff
@@ -293,7 +293,7 @@ def build_sft_dataset(data_cfg, tokenizer, num_samples, answer_only_loss=True, i
             "truncation_method", "right"
         ),  # used to choose truncation method. Options: ['random', 'left', 'right']
         special_tokens=special_tokens,
-        output_original_text=data_cfg.get("output_original_text", False),
+        # output_original_text=data_cfg.get("output_original_text", False),
```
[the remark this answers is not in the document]
trias702: Actually no, output_original_text has been removed from SFTDataset in Nemo upstream

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
trias702: The parent class, MegatronGPTModel still has support for legacy checkpoints in its own `on_load_checkpoint` so I copied that over.

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
trias702: Fixed the comment

examples/nlp/gpt/conf/gpt_spin.yaml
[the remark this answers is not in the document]
trias702: I have updated the comment

nemo_aligner/algorithms/spin.py
```diff
@@ -0,0 +1,504 @@
...
+                global_pbar = tqdm(
+                    self.augment_dataloader(self.train_dataloader),
+                    initial=self.step,
+                    total=self.max_steps,
+                    leave=True,
+                    desc="Training steps",
+                )
+
+                for _, global_batch in zip(loop_iter, global_pbar):
+                    self.model.prepare_for_training()
```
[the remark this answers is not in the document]
trias702: No, it is necessary because doing generation messes up the various MBS/GBS global vars and they need to be reset again. You get crashes if you remove that line

nemo_aligner/algorithms/spin.py
```diff
@@ -0,0 +1,504 @@
...
+        "answers_only": answer_ids,
+        "prompt_lengths": context_lengths,
+        "combined_lengths": combined_lengths,
+    }
+
+    return output
+
+
+class SPINTrainer:
+    """Trainer to coordinate SPIN SFT training
```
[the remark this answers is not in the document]
trias702: Done

nemo_aligner/algorithms/spin.py
```diff
@@ -0,0 +1,504 @@
...
+
+                logprobs = self.model.get_ref_policy_logprobs(new_batch).cpu()
+                act_logps, gen_logps = torch.split(logprobs, len(logprobs) // 2, dim=0)
+
+                new_batch["ref_policy_log_probs_actual"] = act_logps
+                new_batch["ref_policy_log_probs_generated"] = gen_logps
+
+                yield new_batch
+
+                del logprobs, act_logps, gen_logps
```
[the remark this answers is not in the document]
trias702: Done

nemo_aligner/algorithms/spin.py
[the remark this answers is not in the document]
trias702: Done

nemo_aligner/algorithms/spin.py
```diff
@@ -0,0 +1,504 @@
...
+        self.scheduler.step()
+
+        trainer_metrics = {}
+        if grad_norm is not None:
+            trainer_metrics["grad_norm"] = grad_norm
+        trainer_metrics.update({"lr": lr, "loss": loss_mean})
+
+        return loss_mean, {**metrics, **trainer_metrics}
+
+    def _run_inference(self, dataloader_iter, num_microbatches, is_validation):
```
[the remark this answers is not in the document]
trias702: Yup, removed

nemo_aligner/algorithms/spin.py
```diff
@@ -0,0 +1,504 @@
...
+
+        self.length_params = OmegaConf.to_container(self.model.cfg.spin.length_params, resolve=True)
+        self.sampling_params = OmegaConf.to_container(self.model.cfg.spin.sampling_params, resolve=True)
+        self.max_gen_seq_len = self.length_params["max_length"]
+
+    def validation_step(self, global_batch):
+        # these things should go into a GPTModel wrapper
+        self.model.prepare_for_validation_step()
+
+        loss_mean, metrics = self.model.get_loss_and_metrics_vanilla_sft(batch=global_batch, forward_only=True)
```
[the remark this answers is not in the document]
trias702: Done

examples/nlp/gpt/train_gpt_spin.py
```diff
@@ -0,0 +1,186 @@
...
+        eod_mask_loss=cfg.model.data.get("eod_mask_loss", False),
+    )
+
+    train_dataloader = build_dataloader(
+        cfg=cfg,
+        dataset=train_ds,
+        consumed_samples=consumed_samples,
+        mbs=cfg.model.micro_batch_size,
+        gbs=cfg.model.global_batch_size,
+        collate_fn=collate_fn,
```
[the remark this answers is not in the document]
trias702: Correct, yes, good catch, I have changed everything now to be in line with how SFT does it

examples/nlp/gpt/conf/gpt_spin.yaml
```diff
@@ -0,0 +1,182 @@
...
+      # Example of how to specify paths to multiple datasets
+      # file_names:
+      #   - /path/to/squad.jsonl
+      #   - /path/to/mnli.jsonl
+      #   - /path/to/boolq.jsonl
+      # Example of how each dataset is formatted
+      # {'input': 'John von Neumann\nVon Neumann made fundamental contributions .... Q: What did the math of artificial viscosity do?', 'output': 'smoothed the shock transition without sacrificing basic physics'}
+      file_path: ??? # Path to a JSONL file corresponding to the source data. Data format is identical to validation_ds.
+      global_batch_size: ${model.global_batch_size}
+      micro_batch_size: ${model.micro_batch_size}
```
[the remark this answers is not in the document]
trias702: Correct, yes, good catch, I have changed everything now to be in line with how SFT does it

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
trias702: Correct, yes, good catch, I have changed everything now to be in line with how SFT does it

nemo_aligner/algorithms/spin.py
```diff
@@ -0,0 +1,504 @@
...
+                    [
+                        torch.cat(
+                            [
+                                seq,
+                                torch.full((max_batch_len - len(seq),), self.model.tokenizer.eos_id, dtype=seq.dtype),
+                            ]
+                        )
+                        for seq in act_tokens
+                    ]
+                )
```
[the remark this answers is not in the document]
trias702: There is no utility function, I basically took this code as is from PPO, but modified it a little bit

examples/nlp/gpt/conf/gpt_spin.yaml
```diff
@@ -0,0 +1,182 @@
...
+      temperature: 1.0
+      top_k: 0
+      top_p: 1.0
+      repetition_penalty: 1.0
+      add_BOS: False
+      all_probs: False
+      compute_logprob: False
+      # will be used in NeMo version > 1.20.0
+      # keeping it for now
+      end_strings: ["<|endoftext|>", "<extra_id_1>", "\x11"]
```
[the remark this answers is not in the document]
odelalleau: IMO no, actually I think that ideally we would use
```
end_strings: ["<|endoftext|>"]
```
by default, because there is no reason to include in a generic open-source codebase some special tokens that are specific to NVIDIA's internal tokenizers and experiments settings. Especially since this slows down generation when those strings are not encoded as a single token (there was a customer where generation was >2x slower because of this).

The main problem is that unfortunately NeMo included `<extra_id_1>` by default in https://github.com/NVIDIA/NeMo/blob/59aa22d5f65009e8e5bac1d6e8d9bb5be1ec4162/nemo/collections/nlp/modules/common/text_generation_utils.py#L81 so we felt we had to do the same...

nemo_aligner/data/nlp/builders.py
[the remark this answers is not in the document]
odelalleau: Wait, does this mean the current `main` is broken? Or is it a change that should make it to `dev` and not `main`?

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
odelalleau: Yeah NeMo still supports legacy checkpoints but we decided not to support them in NeMo-Aligner. Do you see any reason to make them work with SPIN? (if yes you'll need to be sure this is working, including the loading / saving of the ref. policy).

nemo_aligner/algorithms/spin.py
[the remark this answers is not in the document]
odelalleau: RIght but then can you do it just once after generation instead of on each batch?

examples/nlp/gpt/train_gpt_spin.py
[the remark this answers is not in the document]
odelalleau: As far as I can tell you changed the handling of micro / global batch sizes, but `collate_fn` for the training set is still your `spin_custom_collate` instead of `train_ds.collate_fn`. Is there a specific reason we need your custom function?

nemo_aligner/algorithms/spin.py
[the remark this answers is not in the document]
odelalleau: Can you please create one then? At first glance it seems like this is used in 4 places so it seems worth factorizing (3 times in SPIN, and once at https://github.com/NVIDIA/NeMo-Aligner/blob/6e183b98dbd2798c99bf462e978a43ca90b365d2/nemo_aligner/utils/utils.py#L274)

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
```diff
@@ -400,18 +397,95 @@ def onload_adam_states(self):
...
 
         self.distributed_adam_offload_manager = None
-
-    def sharded_state_dict(self, prefix: str = ""):
-        sharded_state_dict = super().sharded_state_dict(prefix=prefix)
-
-        # add in the reference policy weights
+    
+    # Alternative 1: swap in the ref weights and create a custom key which then needs to be handled in the loader
+    def sharded_state_dict_alt_1(self, prefix: str = ''):
```
odelalleau: Is this working? If yes and you want to keep it, then can you just add a comment to explain that it's not currently used but it's also a valid approach so we keep it around for reference.
(and if not then let's just remove it -- also same thing for `on_load_checkpoint()`)

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
```diff
@@ -400,18 +397,95 @@ def onload_adam_states(self):
...
+                    module_sharded_state_dict = module.sharded_state_dict(prefix=module_prefix)
+                    sharded_state_dict_new.update(module_sharded_state_dict)
+
+            # reset vp rank
+            if parallel_state.get_virtual_pipeline_model_parallel_world_size() is not None:
+                parallel_state.set_virtual_pipeline_model_parallel_rank(0)
+        
+        return sharded_state_dict_orig | sharded_state_dict_new
+    
+    # Alternative 2: use make_sharded_optimizer_tensor to link the TP/PP/DP info, but this may break under VP logic
```
odelalleau, line 442: Maybe raise a `NotImplementedError` if VP is used, to be sure nobody accidentally uses it? (unless you know it crashes already -- I just want to make sure it won't lead to silent incorrect results)

nemo_aligner/utils/utils.py
```diff
@@ -365,3 +368,17 @@ def convert_to_amp_o2_format(state_dict):
         new_state_dict[new_key] = state_dict[key]
 
     return new_state_dict
+
+
+def make_sharded_optimizer_tensor_and_state(model_param, optim_param, prefix: str):
```
odelalleau, line 372: This function looks pretty similar to https://github.com/NVIDIA/Megatron-LM/blob/548e57a00d985c15d5c56fb4749656cf011be9ee/megatron/core/dist_checkpointing/optimizer.py#L66 except that it also handles `ShardedObject`.

Could you please:
1. Add a comment to mention this and explain why we need to handle `ShardedObject` here, so that someone who reads this code has a clue of what this is about.
2. Rename this function and its parameters (and update the error message in the assert above) so that they don't refer to some optimizer? (since we're not using it with an optimizer). They could instead refer to a reference policy but this is a generic function so it could also just be something like `other_param`.

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
```diff
@@ -400,18 +397,95 @@ def onload_adam_states(self):
...
-            sharded_state_dict["reference_policy"] = make_sharded_tensors_for_checkpoint(
-                self.ref_policy_state_dict, prefix
-            )
+            # Link ref_policy keys with sharded_state_dict to reuse sharding information
+            ref_policy_sharded_state_dict = {}
+            for k in self.ref_policy_state_dict:
+                key = k.replace("model.module.", "model.", 1) if self.megatron_amp_O2 else k
+                assert key in sharded_state_dict_orig, f"key [ {key} ] exists in ref_policy but not in sharded_state_dict_orig" # may fail due to nesting?
+                ref_policy_sharded_state_dict[k] = make_sharded_optimizer_tensor_and_state(sharded_state_dict_orig[key], self.ref_policy_state_dict[k], 'reference_policy')
+            sharded_state_dict_orig['reference_policy'] = ref_policy_sharded_state_dict
```
odelalleau: Related to your comment above about nesting:

I wonder if there's a risk of some code breaking (in the future) when we use nested sharded state dicts. Would it be safer to keep it flat and instead rely on the `reference_policy` prefix to (i) prevent overlap with current params, and (ii) reconstruct the ref policy state dict in `on_load_checkpoint()`?

odelalleau (commented):
Congrats on getting the ref policy sharded state dict stuff working! It looks like this really wasn't trivial => a few clarifying questions / comments below.

odelalleau:
Oh, also just a note that we'll need CI as well as documentation for this new algorithm. I don't remember exactly how to do these so please synch with @gshennvm for this.

examples/nlp/gpt/conf/gpt_spin.yaml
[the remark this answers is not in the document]
trias702: Ah interesting, I didn't know this, okay, I will remove it

examples/nlp/gpt/train_gpt_spin.py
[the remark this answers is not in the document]
trias702: Yes, for SPIN, we need custom output from each batch coming from the dataloader which `train_ds.collate_fn` does not give us. These custom outputs are necessary for SPIN to work correctly, hence we need `spin_custom_collate`. You can look at both functions to see the differences between them

nemo_aligner/algorithms/spin.py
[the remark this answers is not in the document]
trias702: No, because look at how `global_pbar` is constructed:

```
global_pbar = tqdm(
                    self.augment_dataloader(self.train_dataloader),
                    initial=self.step,
                    total=self.max_steps,
                    leave=True,
                    desc="Training steps",
                )
```

We wrap the train_dataloader with the augment function, which means for every iteration of the for-loop, `augment_dataloader` will fire, and `augment_dataloader` includes the generation which resets the global MBS/GBS values, hence we need to reset them after the for-loop fires but before `train_single_step`

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
trias702: Yes, it works correctly, I will add a comment explaining that it is

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
trias702: Okay, no worries, I have removed the legacy checkpoint code

nemo_aligner/data/nlp/builders.py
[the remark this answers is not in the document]
odelalleau: Ok makes sense, let's keep it then since @gshennvm is about to update to 1.23 in #116

examples/nlp/gpt/train_gpt_spin.py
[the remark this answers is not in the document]
odelalleau: Got it, thanks. I had noticed they were quite different and originally it wasn't obvious to me why, especially since you were using `validation_ds.collate_fn`. But I get it now -- validation is very different from training in your SPIN implementation since it's using an SFT loss, which explains this discrepancy.

nemo_aligner/algorithms/spin.py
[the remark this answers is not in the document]
odelalleau: Got it, thanks, I forgot generation was happening on each batch.

nemo_aligner/data/nlp/builders.py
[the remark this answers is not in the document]
trias702: Once he pushes out 116, I'll update here and uncomment that line

nemo_aligner/algorithms/spin.py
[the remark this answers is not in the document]
trias702: Added!

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
trias702: Added

nemo_aligner/utils/utils.py
trias702: Done

trias702:
> Oh, also just a note that we'll need CI as well as documentation for this new algorithm. I don't remember exactly how to do these so please synch with @gshennvm for this.

The CI will come after the PR is merged because it needs to get merged for testing the new CI components. I will work on the documentation this week.

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
trias702: Moving to nesting may indeed cause several breaking changes, not just here, but with loads of other code in Aligner (for example `retrieve_model_state_dict_in_cpu`. I think if we move to nested state_dicts we will need to change a lot and run several tests. Ideally, everything would remain flat, but that unfortunately isn't possible if using VP

nemo_aligner/data/nlp/builders.py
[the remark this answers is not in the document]
odelalleau: Actually I just double checked as I was surprised that our `main` would be broken, and you can uncomment now: you're probably not using the latest Dockerfile which cherry-picks https://github.com/NVIDIA/NeMo/commit/7d3d9ac3b1aecf5786b5978a0c1e574701473c62

nemo_aligner/utils/utils.py
odelalleau: Thanks -- sorry to be annoying, but I'm afraid the way you named parameters is confusing because `reference_param` is actually not used for the parameters of the reference policy. As far as I can understand the code is correct and it's just that "reference" may actually refer to different concepts, which is likely to confuse people looking at this code. Could you please rename again? (for instance you could use "base" or "template" instead of "reference")

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
odelalleau: PS: I just had a quick look at your `alt_1` implementation (which I had skipped earlier as I wasn't sure it was supposed to work), and actually what I describe (in terms of keeping the dict flat) seems to be exactly what you did there.

examples/nlp/gpt/conf/gpt_spin.yaml
```diff
@@ -0,0 +1,179 @@
...
+  precision: bf16-mixed
+  
+  # spin specific args
+  spin:
+    max_iterations: 3   # the number of iterations (rounds) for which the SPIN "game" is played. This is T in the original paper
+    max_epochs: 1       # the number of epochs to use per iteration. In the paper, they do 3 iterations with 2 epochs per iteration
+    max_steps: -1
+    val_check_interval: 0.1
+    save_interval: 100
+    limit_steps_per_epoch: -1
```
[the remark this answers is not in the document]
odelalleau: Actually, thinking about it a bit more I think this won't work because the sampler won't account for it and thus its internal epoch count will be incorrect, right?
Maybe we can remove this option for now and add it back later? (I also want to make some changes to the sampler so it may be best to wait before modifying it to support this)

nemo_aligner/data/nlp/builders.py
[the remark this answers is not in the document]
trias702: Cheers, I'll uncomment for the final commit

nemo_aligner/utils/utils.py
trias702: Sorry if it's confusing, but the parameter names now are correct to the logic of what is actually occuring in the function. We read the TP/PP/DP axis information from the `reference_param` and use that information so save down (via dataclasses.replace) the `model_param`. This is what the function does and how it used to save the ref policy in SPIN. The reference_param input is not being changed, it is being used as a reference for the TP/PP/DP axis information in order to correctly save in sharded format. If you don't have a reference object to read this from, then your model won't save the correct TP/PP/DP axis.

Ergo, the function now has correct param names in the spirit of what it is actually doing. It has nothing to do with the reference policy in SPIN, that's just one algo which uses this function.

examples/nlp/gpt/conf/gpt_spin.yaml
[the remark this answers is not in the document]
trias702: Good point, the sampler will have an incorrect internal epoch count given the way I have written it. We would need to have an accessor method on the sampler itself which allows us to internally change its `total_samples` however, I'm not sure if that would work because the core dataloader would still be wrong in terms of `__len__`. Will need to think on this some more.

In the meantime, this can be accomplished manually by just changing the input dataset size, but that is annoying to need to do.

nemo_aligner/algorithms/spin.py
```diff
@@ -0,0 +1,467 @@
...
+            self.model.ref_policy_state_dict = None
+
+        self.ckpt_callback.custom_save(monitor_candidates=monitor_candidates, is_train_end=is_train_end)
+
+        self.model.finish_training()
+
+    def set_max_steps(self):
+        self.max_steps = self.num_steps_per_epoch * self.cfg.max_epochs * self.cfg.max_iterations
+
+        if (max_steps := self.cfg.get("max_steps", -1)) > 0:
```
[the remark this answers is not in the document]
trias702: I couldn't think of a reason why max_steps = 0 would ever be used, so I changed it, but I guess your point sort of makes sense if you want to do validation only. I'll change it back.

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
trias702: `f"{prefix}reference_policy"`

This is already happening inside of `utils.make_sharded_tensors_from_reference`. If you look at the code of that function, and the fact that I pass in "reference_policy" you will see that the key it uses is what you propose.

Using Alt_2 as it is now, I can see that everything is saved down correctly when I look in the root of the checkpoint directory:

```
drwxr-xr-x  2 dip dip 4.0K Feb 29 19:00 model.decoder.final_layernorm.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 18:59 model.decoder.layers.mlp.linear_fc1._extra_state
drwxr-xr-x  2 dip dip 4.0K Feb 29 19:00 model.decoder.layers.mlp.linear_fc1.layer_norm_weight
drwxr-xr-x  2 dip dip  20K Feb 29 19:00 model.decoder.layers.mlp.linear_fc1.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 18:59 model.decoder.layers.mlp.linear_fc2._extra_state
drwxr-xr-x  2 dip dip  12K Feb 29 19:00 model.decoder.layers.mlp.linear_fc2.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 18:59 model.decoder.layers.self_attention.linear_proj._extra_state
drwxr-xr-x  2 dip dip  12K Feb 29 19:00 model.decoder.layers.self_attention.linear_proj.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 18:59 model.decoder.layers.self_attention.linear_qkv._extra_state
drwxr-xr-x  2 dip dip 4.0K Feb 29 19:00 model.decoder.layers.self_attention.linear_qkv.layer_norm_weight
drwxr-xr-x  2 dip dip  12K Feb 29 19:00 model.decoder.layers.self_attention.linear_qkv.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 18:59 model.embedding.word_embeddings.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 19:00 model.output_layer.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 18:59 optimizer.state.exp_avg.model.decoder.final_layernorm.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 18:59 optimizer.state.exp_avg.model.decoder.layers.mlp.linear_fc1.layer_norm_weight
drwxr-xr-x  2 dip dip  20K Feb 29 18:59 optimizer.state.exp_avg.model.decoder.layers.mlp.linear_fc1.weight
drwxr-xr-x  2 dip dip  12K Feb 29 18:59 optimizer.state.exp_avg.model.decoder.layers.mlp.linear_fc2.weight
drwxr-xr-x  2 dip dip  12K Feb 29 18:59 optimizer.state.exp_avg.model.decoder.layers.self_attention.linear_proj.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 18:59 optimizer.state.exp_avg.model.decoder.layers.self_attention.linear_qkv.layer_norm_weight
drwxr-xr-x  2 dip dip  12K Feb 29 18:59 optimizer.state.exp_avg.model.decoder.layers.self_attention.linear_qkv.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 18:59 optimizer.state.exp_avg.model.embedding.word_embeddings.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 18:59 optimizer.state.exp_avg.model.output_layer.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 18:59 optimizer.state.exp_avg_sq.model.decoder.final_layernorm.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 18:59 optimizer.state.exp_avg_sq.model.decoder.layers.mlp.linear_fc1.layer_norm_weight
drwxr-xr-x  2 dip dip  16K Feb 29 18:59 optimizer.state.exp_avg_sq.model.decoder.layers.mlp.linear_fc1.weight
drwxr-xr-x  2 dip dip  12K Feb 29 18:59 optimizer.state.exp_avg_sq.model.decoder.layers.mlp.linear_fc2.weight
drwxr-xr-x  2 dip dip  12K Feb 29 18:59 optimizer.state.exp_avg_sq.model.decoder.layers.self_attention.linear_proj.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 18:59 optimizer.state.exp_avg_sq.model.decoder.layers.self_attention.linear_qkv.layer_norm_weight
drwxr-xr-x  2 dip dip  12K Feb 29 18:59 optimizer.state.exp_avg_sq.model.decoder.layers.self_attention.linear_qkv.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 18:59 optimizer.state.exp_avg_sq.model.embedding.word_embeddings.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 18:59 optimizer.state.exp_avg_sq.model.output_layer.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 18:59 optimizer.state.fp32_param.model.decoder.final_layernorm.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 18:59 optimizer.state.fp32_param.model.decoder.layers.mlp.linear_fc1.layer_norm_weight
drwxr-xr-x  2 dip dip  12K Feb 29 18:59 optimizer.state.fp32_param.model.decoder.layers.mlp.linear_fc2.weight
drwxr-xr-x  2 dip dip  12K Feb 29 18:59 optimizer.state.fp32_param.model.decoder.layers.self_attention.linear_proj.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 18:59 optimizer.state.fp32_param.model.decoder.layers.self_attention.linear_qkv.layer_norm_weight
drwxr-xr-x  2 dip dip  12K Feb 29 18:59 optimizer.state.fp32_param.model.decoder.layers.self_attention.linear_qkv.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 18:59 optimizer.state.fp32_param.model.embedding.word_embeddings.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 18:59 optimizer.state.fp32_param.model.output_layer.weight
drwxr-xr-x  2 dip dip  16K Feb 29 18:59 optimizer.state.param.model.decoder.layers.mlp.linear_fc1.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 19:00 reference_policy.model.decoder.final_layernorm.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 18:59 reference_policy.model.decoder.layers.mlp.linear_fc1._extra_state
drwxr-xr-x  2 dip dip 4.0K Feb 29 19:00 reference_policy.model.decoder.layers.mlp.linear_fc1.layer_norm_weight
drwxr-xr-x  2 dip dip  20K Feb 29 19:00 reference_policy.model.decoder.layers.mlp.linear_fc1.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 18:59 reference_policy.model.decoder.layers.mlp.linear_fc2._extra_state
drwxr-xr-x  2 dip dip  12K Feb 29 19:00 reference_policy.model.decoder.layers.mlp.linear_fc2.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 18:59 reference_policy.model.decoder.layers.self_attention.linear_proj._extra_state
drwxr-xr-x  2 dip dip  12K Feb 29 19:00 reference_policy.model.decoder.layers.self_attention.linear_proj.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 18:59 reference_policy.model.decoder.layers.self_attention.linear_qkv._extra_state
drwxr-xr-x  2 dip dip 4.0K Feb 29 19:00 reference_policy.model.decoder.layers.self_attention.linear_qkv.layer_norm_weight
drwxr-xr-x  2 dip dip  12K Feb 29 19:00 reference_policy.model.decoder.layers.self_attention.linear_qkv.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 19:00 reference_policy.model.embedding.word_embeddings.weight
drwxr-xr-x  2 dip dip 4.0K Feb 29 19:00 reference_policy.model.output_layer.weight
```

I understand what you mean by flat now though, and I'm honestly not sure if that is necessary. I think it's okay to leave it for now and then engineer for confirmed cases in the future rather than trying to over-engineer now for possible future cases which may never come to fruition.

examples/nlp/gpt/conf/gpt_spin.yaml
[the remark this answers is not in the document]
trias702: I have changed the parameter name and default value to be in line with PTL. It works fine as is, but we'd need to make some changes to the Sampler to have it behave as PTL intended

trias702:
Documentation has been added

nemo_aligner/utils/utils.py
odelalleau: Yeah, I understand that, I'm just saying this is quite confusing to someone reading the code and I was thus suggesting to rename the parameters to avoid this confusion.

If you insist on keeping the current names can you please at least rephrase this comment
```
# (...) can be copied from some
# reference model, like in the case of a reference policy which has a 1:1 architecture as some actor policy
```
which reinforces this confusion since the "reference model" is not actually the "reference policy" in practice.

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
odelalleau: Ok, I'm not familiar enough with this distributed checkpointing code to know for sure what is the "right" way of doing things, so let's keep your current implementation for now since it works.

examples/nlp/gpt/conf/gpt_spin.yaml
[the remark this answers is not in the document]
odelalleau: Ok I'll look into fixing it when I do these sampler changes.

But I think this is going to crash with `limit_train_batches=null`:
```python
if (limit_train_batches := self.cfg.get("limit_train_batches", -1)) > 0:
```
=> should be instead
```python
if (limit_train_batches := self.cfg.get("limit_train_batches")) is not None:
```

docs/user-guide/SPIN.rst
```diff
@@ -0,0 +1,181 @@
...
+However, you can also run SPIN on models which have been SFTed on instruction-based datasets as well, similar to DPO/PPO. Either type of model will work well with SPIN. If you would like to start with a supervised fine tuned model instead of a base model, please see our full guide on how to perform SFT on a Megatron GPT model :ref:`SFT guide <sft>`.
+
+SPIN Model Training
+###################
+
+SPIN training uses the exact same dataset formatting and files as the NeMo-Aligner SFT trainer. Please see the data formatting section of SFT to understand the data format necessary for SPIN :ref:`SFT guide <sft>`
+
+Once your data is processed into the correct format you are ready to begin SPIN training. You must start with a pretrained or SFT trained model. For this section we will use the SFT model trained in the previous step to train the SPIN model.
+For the purposes of the following sections, we'll assuming your training jsonl file is located in ``/path/to/train_spin_format.jsonl`` and your validation jsonl file is located in ``/path/to/valid_spin_format.jsonl``.
```
odelalleau, line 69: Small typo
```suggestion
For the purposes of the following sections, we'll assume your training jsonl file is located in ``/path/to/train_spin_format.jsonl`` and your validation jsonl file is located in ``/path/to/valid_spin_format.jsonl``.
```

docs/user-guide/SPIN.rst
```diff
@@ -0,0 +1,181 @@
...
+
+For the below parameters, the ``model.spin.ref_policy_kl_penalty`` corresponds to the beta parameter in the SPIN paper, and ``trainer.spin.max_iterations`` corresponds to T (with ``trainer.spin.max_epochs`` per iteration)
```
odelalleau, line 71: Missing "epochs"
```suggestion
For the below parameters, the ``model.spin.ref_policy_kl_penalty`` corresponds to the beta parameter in the SPIN paper, and ``trainer.spin.max_iterations`` corresponds to T (with ``trainer.spin.max_epochs`` epochs per iteration)
```

examples/nlp/gpt/conf/gpt_spin.yaml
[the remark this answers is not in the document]
trias702: Doesn't making it null in the config yaml mean that when you call .get on self.cfg, it returns a None which causes it to populate the -1 as the default value?

docs/user-guide/SPIN.rst
```diff
@@ -0,0 +1,181 @@
...
+         .. code-block:: bash 
+
+            export GPFS="/path/to/nemo-aligner-repo"
+            export TRAIN_DATA_PATH="/path/to/train_spin_format.jsonl"
+            export VALID_DATA_PATH="/path/to/valid_spin_format.jsonl"
+
+            python -u ${GPFS}/examples/nlp/gpt/train_gpt_spin.py \
+               trainer.num_nodes=1 \
+               trainer.devices=8 \
+               ++model.mcore_gpt=true \
```
odelalleau, line 89: This line can be removed (already in the default config and should never be changed)

docs/user-guide/SPIN.rst
```diff
@@ -0,0 +1,181 @@
...
+               "model.data.validation_ds.file_path=${VALID_DATA_PATH}" \
+               exp_manager.create_wandb_logger=false \
+               exp_manager.wandb_logger_kwargs.project=spin_training \
+               exp_manager.wandb_logger_kwargs.name=spin_training \
+               exp_manager.explicit_log_dir=/results \
+               ++trainer.spin.max_iterations=1 \
+               ++trainer.spin.max_epochs=1 \
+               ++model.spin.ref_policy_kl_penalty=0.1
+               ++model.spin.length_params.max_length=2048 \
+               ++model.data.train_ds.max_seq_length=4096 \
```
odelalleau, line 103: Couple of comments:
1. Small syntax error probably due to moving lines around => the line with `ref_policy_kl_penalty` should end with a `\` and the last line shouldn't
2. Better remove the `++` in front of all of these options since they are in our config so the `++` is useless (otherwise it's confusing as to why we need `++`, and also we may not notice that our tutorials are broken if a variable is renamed)

docs/user-guide/SPIN.rst
```diff
@@ -0,0 +1,181 @@
...
+
+    .. tab-item:: Slurm
+        :sync: key4
+
+         To run SPIN model training using Slurm. The script below uses 32 nodes, but you can change the node count to something smaller.
```
odelalleau, line 108: Any reason to use such a high number of nodes? I'm afraid people may get scared that our implementation requires 256 GPUs just to train a small (2B / 7B) model.

docs/user-guide/SPIN.rst
```diff
@@ -0,0 +1,181 @@
...
+            mkdir -p ${RESULTS_DIR}
+
+            read -r -d '' cmd <<EOF
+            echo "*******STARTING********" \
+            && echo "---------------" \
+            && echo "Starting training" \
+            && cd ${GPFS} \
+            && export PYTHONPATH="${GPFS}:${PYTHONPATH}" \
+            && export HYDRA_FULL_ERROR=1 \
+            && python -u ${GPFS}/examples/nlp/gpt/train_gpt_dpo.py \
```
odelalleau, line 147: Missed one ;)
```suggestion
            && python -u ${GPFS}/examples/nlp/gpt/train_gpt_spin.py \
```

docs/user-guide/SPIN.rst
```diff
@@ -0,0 +1,181 @@
...
+               ++model.global_batch_size=64 \
+               exp_manager.explicit_log_dir=${RESULTS_DIR} \
+               exp_manager.create_wandb_logger=True \
+               exp_manager.wandb_logger_kwargs.name=${NAME} \
+               exp_manager.wandb_logger_kwargs.project=${PROJECT} \
+               ++trainer.spin.max_iterations=1 \
+               ++trainer.spin.max_epochs=1 \
+               ++model.spin.ref_policy_kl_penalty=0.1
+               ++model.spin.length_params.max_length=2048 \
+               ++model.data.train_ds.max_seq_length=4096 \
```
odelalleau, line 163: Same two comments as above (on `\` and `++`)

docs/user-guide/SPIN.rst
```diff
@@ -0,0 +1,181 @@
...
+            EOF
+
+            srun -o $OUTFILE -e $ERRFILE --container-image=$CONTAINER $MOUNTS bash -c "${cmd}"
+            set +x
+
+During SPIN training, there will be several metrics recorded to WandB which you can monitor, chiefly acc (representing the percentage amount whereby the model's actual [ground truth label] rewards are greater than the generated rewards [generated by the reference policy]).
```
odelalleau, line 169: This may be a bit confusing because the reference policy doesn't generate rewards.

What about for instance
```suggestion
During SPIN training, there will be several metrics recorded to WandB which you can monitor, chiefly acc (representing the percentage amount whereby the model's implicit reward for the ground truth response is greater than for the response generated by the reference policy).
```

docs/user-guide/SPIN.rst
```diff
@@ -0,0 +1,181 @@
...
+The ``reward`` in this case is calculated as the difference between model log probs and the reference log probs, multiplied by the KL penalty (beta in the original paper), for the ground truth and generated responses.
+During training, the acc should generally be increasing, but don't worry if its absolute value remains low, as it doesn't correlate to finalised MTBench or MMLU scores. It should just be generally increasing.
+Other metrics to keep an eye on are the rewards_actual_mean and rewards_generated_mean, which represent the average of the ``rewards`` as defined above. Again, the absolute values aren't necessarily so important as the fact that the actual_mean should be greater than the generated_mean over time, and the greater that difference, the better.
+All metrics will be grouped by either ``train/`` or ``val/`` in WandB, representing whether that metric is from the training or validation set, respectively.
```
odelalleau, line 173: Maybe also mention that we don't generate responses for validation so there is no accuracy metric? (but instead it's a standard SFT NLL loss)

docs/user-guide/SPIN.rst
```diff
@@ -0,0 +1,181 @@
...
+
+When it comes to ideal hyperparameters for SPIN training, much will depend on the characteristics of your SFT (or base/foundational) model and your training data, so there is no one-size-fits-all parameter set which will work in all cases.
+However, the following the following is a brief overview of which hyperparameters we have perturbed for various model sizes and their effects:
```
odelalleau, line 176:
```suggestion
However, the following is a brief overview of which hyperparameters we have perturbed for various model sizes and their effects:
```

docs/user-guide/SPIN.rst
```diff
@@ -0,0 +1,181 @@
...
+
+* global_batch_size: the SPIN paper recommends 64 for a 7B model, which we have found holds true, in that higher GBS for 7B models performs much worse. For larger models, you can increase to 128 or 256 as needed, but recommend you start with 64 as a baseline
```
odelalleau, line 178: add "we"
```suggestion
* global_batch_size: the SPIN paper recommends 64 for a 7B model, which we have found holds true, in that higher GBS for 7B models performs much worse. For larger models, you can increase to 128 or 256 as needed, but we recommend you start with 64 as a baseline
```

docs/user-guide/SPIN.rst
```diff
@@ -0,0 +1,181 @@
...
+* iterations/epochs: the SPIN paper used iterations=3 and epochs=2 for their training on a 7B model with a training dataset size of 200k. Using the same foundational model as the authors, we found better results with iterations=1, epochs=1 using a 50k subset of their 200k data. We therefore recommend starting with iterations=1, and increasing to 2 as needed by testing on MT-Bench/MMLU.
```
odelalleau, line 179: I think we should mention the fact that our current implementation doesn't increase the size of the training set on iterations 2+, as in the SPIN paper. We don't know to which extent this may be important to get good results with more iterations (right?).

odelalleau (changes requested):
Thanks for writing the doc! Just a few small comments (mostly typos)

examples/nlp/gpt/conf/gpt_spin.yaml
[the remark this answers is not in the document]
odelalleau: No, `.get()` on a dict only returns the default value if the key is missing:
```
In [14]: {"foo": None}.get("foo", -1) is None
Out[14]: True
```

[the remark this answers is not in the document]
trias702: Okay, but we also need to check that the value is > 0, because someone can just change the null to -1 and it would pass your if check. Can we do something like:

```
if (limit_train_batches := self.cfg.get("limit_train_batches")) is not None and limit_train_batches > 0:
            self.num_steps_per_epoch = min(self.num_steps_per_epoch, limit_train_batches)
```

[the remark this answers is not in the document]
odelalleau: From what I can tell from the PTL docs `limit_train_batches < 0` isn't allowed, that being said someone may mistakenly use `-1` instead of `None` so maybe you can just add below the `if`:
```python
assert limit_train_batches >= 0, "`limit_train_batches` must be either a non-negative integer or `None`"
```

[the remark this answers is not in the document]
trias702: It would be nice to keep it to a one-liner, what do you think of what I have checked in just now?

docs/user-guide/SPIN.rst
trias702: Fixed, thanks

docs/user-guide/SPIN.rst
trias702: Fixed, thanks

docs/user-guide/SPIN.rst
trias702: Fixed, thanks

docs/user-guide/SPIN.rst
trias702: Fixed, thanks

docs/user-guide/SPIN.rst
trias702: Fixed, thanks

docs/user-guide/SPIN.rst
trias702: I changed, it thanks for the suggestion

docs/user-guide/SPIN.rst
trias702: Fixed, thanks

docs/user-guide/SPIN.rst
trias702: Fixed, thanks

docs/user-guide/SPIN.rst
trias702: I have added a section discussing this, let me know if it suffices

docs/user-guide/SPIN.rst
trias702: I have added a line about that, let me know if it suffices

docs/user-guide/SPIN.rst
trias702: We currently use 32 nodes for a 7B model because generation takes so much time, so I carried it across. I have since changed it to 4 nodes in the example (since we're using a 2B model instead of 7B)

docs/user-guide/SPIN.rst
trias702: Fixed, thanks

examples/nlp/gpt/conf/gpt_spin.yaml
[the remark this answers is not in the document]
odelalleau: I have a slight preference for crashing to stick to PTL's API, but it's not a huge deal.

That being said though I took a quick look at how PTL is using it and it looks like a value of 0 is allowed to skip training so it may be better to use `>=` instead of `>`: https://github.com/Lightning-AI/lightning/blob/34a34a0754d612965b158f1f588336507554fa16/src/lightning/pytorch/loops/fit_loop.py#L194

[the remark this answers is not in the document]
trias702: This is fixed now

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
```diff
@@ -0,0 +1,632 @@
...
+                "avoid explicit casting for pipeline communication"
+            )
+
+        self.ref_policy_state_dict = None
+        self.distributed_adam_offload_manager = None
+        self.to_offload_adam_states = self.cfg.spin.offload_adam_states
+
+        # self.ref_policy_kl_penalty = self.cfg.spin.get("ref_policy_kl_penalty", 0.0)
+        self.spin_config = OmegaConf.to_container(self.cfg.spin, resolve=True)
+        if isinstance(self.spin_config["ref_policy_kl_penalty"], float):
```
odelalleau, line 77: Safer as some people may use `0` in the config:
```suggestion
        if isinstance(self.spin_config["ref_policy_kl_penalty"], (float, int)):
```

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
```diff
@@ -0,0 +1,632 @@
...
+                "when using pipeline parallelism, it is recommended to set megatron_amp_O2 to be True to "
+                "avoid explicit casting for pipeline communication"
+            )
+
+        self.ref_policy_state_dict = None
+        self.distributed_adam_offload_manager = None
+        self.to_offload_adam_states = self.cfg.spin.offload_adam_states
+
+        # self.ref_policy_kl_penalty = self.cfg.spin.get("ref_policy_kl_penalty", 0.0)
+        self.spin_config = OmegaConf.to_container(self.cfg.spin, resolve=True)
```
odelalleau, line 76: Not a big deal but just a note that the code could be simplified a bit by storing `self._ref_policy_kl_penalty = self.cfg.spin.ref_policy_kl_penalty` (instead of keeping around `self.spin_config`), since we never access anything else inside `self.spin_config`.

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
```diff
@@ -0,0 +1,632 @@
...
+        if isinstance(self.spin_config["ref_policy_kl_penalty"], float):
+            self.ref_policy_kl_penalty = self.spin_config["ref_policy_kl_penalty"]
+        else:
+            self.ref_policy_kl_penalty = 0.0
```
odelalleau, line 80: Can you assert that it's a list here? (or raise a an explicit `TypeError`)

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
```diff
@@ -0,0 +1,632 @@
...
+        pos_ids = torch.cat((batch["position_ids"], batch["position_ids"]), dim=0)
+        global_batch = [tokens, masks, pos_ids]
+        with cpu_weight_swap(self, self.ref_policy_state_dict, megatron_amp_O2=self.megatron_amp_O2):
+            ref_log_probs = self.get_logprob_batch(global_batch)
+
+        # return in GPU, trainer needs to move to cpu
+        return ref_log_probs
+
+    def set_KL_penalty_by_iteration(self, iteration):
+        if isinstance(self.spin_config["ref_policy_kl_penalty"], float):
```
odelalleau, line 625:
```suggestion
        if isinstance(self.spin_config["ref_policy_kl_penalty"], (float, int)):
```

examples/nlp/gpt/conf/gpt_spin.yaml
```diff
@@ -0,0 +1,179 @@
...
+
+model:  
+  mcore_gpt: True
+  micro_batch_size: 1
+  global_batch_size: 64
+  megatron_amp_O2: True
+
+  spin:
+    log_prob_forward_micro_batch_size: 1
+    ref_policy_kl_penalty: 0.2
```
odelalleau, line 64: Please add a comment to document the schedule

odelalleau (changes requested):
A few small requests for the new schedule

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
trias702: Yeah I considered doing that, but figured I'd save the whole config in case we needed to access anything else down the line. I'm pretty ambivalent about doing it the current way or just saving the KL part, if you feel strongly one way or another let me know.

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
trias702: Fixed, thanks

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
trias702: I added a raise TypeError

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
trias702: Fixed, thanks

examples/nlp/gpt/conf/gpt_spin.yaml
trias702: Done

nemo_aligner/models/nlp/gpt/megatron_gpt_spin_model.py
odelalleau: No strong feeling

odelalleau (approved):
Good to go as far as I'm concerned, thanks for this worthy addition!

gshennvm (approved):
thanks for the change! i will F/U on the CI

gshennvm (dismissed):

odelalleau (approved):

odelalleau (approved):

gshennvm (approved):

Issue #107, opened 2024-02-14
random samplers keeps state


gshennvm:
since #73 we've switched to using random samplers to support multiple epochs. 

However, the `__iter__` method of the random sampler keeps [state](https://github.com/NVIDIA/NeMo-Aligner/blob/8399e5af0072c2a12fc9ec7f68a7051bb95e374a/nemo_aligner/data/nlp/samplers.py#L97) which causes validation on a subset of our data as in PPO to use a **different** subset each time.

In PPO we try to reset the optimizer by calling `[__iter__](https://github.com/NVIDIA/NeMo-Aligner/blob/8399e5af0072c2a12fc9ec7f68a7051bb95e374a/nemo_aligner/algorithms/ppo.py#L291)` on it, but because the random sampler keeps state, it does not actually reset the validation set

Issue #27, opened 2023-12-01
PPO is missing `limit_val_batches`


odelalleau:
**Is your feature request related to a problem? Please describe.**

The PPO trainer should support `limit_val_batches`, like DPO / RM / SFT.

PR #112, opened 2024-02-26
Fix random sampler keeping state with validation runs

gleibovich-nvidia:
Fixes issue #107 by allowing having `build_dataloader` support both deterministic and random samplers, and revert PPO `val_dataloader` to be a deterministic one, thus allowing tracking the same batch during a PPO run. 

The same issue seems to also be present with DPO. Is the expected behavior same as with PPO and we'd like to use a deterministic batch sampler with DPO as well?

### DIFF
--- a/examples/nlp/gpt/train_gpt_ppo_actor.py
+++ b/examples/nlp/gpt/train_gpt_ppo_actor.py
@@ -130,7 +130,7 @@ def main(cfg) -> None:
         gbs=cfg.model.ppo.num_val_samples,
         collate_fn=collate_fn,
         load_gbs=False,
-        use_random_sampler=False
+        use_random_sampler=False,
     )
 
     # nemo uses the train dataloader to figure out
--- a/examples/nlp/gpt/train_gpt_ppo_actor.py
+++ b/examples/nlp/gpt/train_gpt_ppo_actor.py
@@ -130,6 +130,7 @@ def main(cfg) -> None:
         gbs=cfg.model.ppo.num_val_samples,
         collate_fn=collate_fn,
         load_gbs=False,
+        use_random_sampler=False
     )
 
     # nemo uses the train dataloader to figure out
--- a/nemo_aligner/data/nlp/builders.py
+++ b/nemo_aligner/data/nlp/builders.py
@@ -342,14 +342,12 @@ def build_dataloader(
     if hasattr(cfg.model.data, "dataloader_type") and cfg.model.data.dataloader_type == "single":
         if use_random_sampler:
             cls = MegatronPretrainingRandomBatchSampler if load_gbs else MegatronPretrainingRandomSampler
-            common_params['seed'] = cfg.model.seed
+            common_params["seed"] = cfg.model.seed
         else:
             cls = MegatronPretrainingBatchSampler if load_gbs else MegatronPretrainingSampler
         batch_sampler = cls(**common_params)
     else:
-        raise ValueError(
-            '`cfg.model.data.dataloader_type` must be set to "single"'
-        )
+        raise ValueError('`cfg.model.data.dataloader_type` must be set to "single"')
 
     return torch.utils.data.DataLoader(
         dataset,
--- a/nemo_aligner/data/nlp/builders.py
+++ b/nemo_aligner/data/nlp/builders.py
@@ -352,7 +352,7 @@ def build_dataloader(
         batch_sampler = create_batch_sampler(cls, **common_params, seed=cfg.model.seed if use_random_sampler else None)
     else:
         raise ValueError(
-            'cfg.model.data.dataloader_type must be "single" or cfg.model.data.dataloader_type not found.'
+            'cfg.model.data.dataloader_type` must be set to "single"
         )
 
     return torch.utils.data.DataLoader(
--- a/nemo_aligner/data/nlp/builders.py
+++ b/nemo_aligner/data/nlp/builders.py
@@ -321,7 +321,7 @@ def build_dataloader(
     pad_samples_to_global_batch_size=False,
     collate_fn=None,
     load_gbs=True,
-    use_random_sampler=True
+    use_random_sampler=True,
 ):
     """Buld dataloader given an input dataset."""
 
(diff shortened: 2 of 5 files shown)


nemo_aligner/data/nlp/builders.py
```diff
@@ -321,29 +321,39 @@ def build_dataloader(
...
+        "total_samples": len(dataset),
+        "consumed_samples": consumed_samples,
+        "micro_batch_size": mbs,
+        "data_parallel_rank": parallel_state.get_data_parallel_rank(),
+        "data_parallel_size": parallel_state.get_data_parallel_world_size(),
+        "drop_last": drop_last,
+        "global_batch_size": gbs,
+        "pad_samples_to_global_batch_size": pad_samples_to_global_batch_size,
+    }
+
```
gshennvm, line 344: can you add a if `use_random_sampler` and then put the seed into the common_params.

this way we don't need the `create_batch_sampler` function

examples/nlp/gpt/train_gpt_ppo_actor.py
```diff
@@ -130,6 +130,7 @@ def main(cfg) -> None:
         gbs=cfg.model.ppo.num_val_samples,
         collate_fn=collate_fn,
         load_gbs=False,
+        use_random_sampler=False,
```
gshennvm, line 133: i think we should also reflect this change on RM, SFT and DPO

gshennvm (changes requested):
thanks for working on this! 

make sure to also update the CHANGELOG

nemo_aligner/data/nlp/builders.py
```diff
@@ -321,29 +321,39 @@ def build_dataloader(
...
-            )
         else:
-            raise ValueError('cfg.data.dataloader_type must be "single"')
+            cls = MegatronPretrainingSampler
+
+        batch_sampler = create_batch_sampler(cls, **common_params, seed=cfg.model.seed if use_random_sampler else None)
     else:
-        raise ValueError('cfg.data.dataloader_type not found. Must be "single"')
+        raise ValueError(
+            'cfg.model.data.dataloader_type must be "single" or cfg.model.data.dataloader_type not found.'
```
odelalleau, line 355: This error message seems confusing to me. Suggestion:
```suggestion
            'cfg.model.data.dataloader_type` must be set to "single"
```

nemo_aligner/data/nlp/builders.py
```diff
@@ -321,29 +321,39 @@ def build_dataloader(
...
-                data_parallel_rank=parallel_state.get_data_parallel_rank(),
-                data_parallel_size=parallel_state.get_data_parallel_world_size(),
-                drop_last=drop_last,
-                global_batch_size=gbs,
-                pad_samples_to_global_batch_size=pad_samples_to_global_batch_size,
-                seed=cfg.model.seed,
-            )
         else:
-            raise ValueError('cfg.data.dataloader_type must be "single"')
+            cls = MegatronPretrainingSampler
```
odelalleau, line 350: I'm a bit suspicious of `load_gbs` not being taken into account here. Does this work with `load_gbs=True`?

odelalleau, line 350: I actually believe we should shuffle the validation set by default. This is safer when using `limit_val_batches` (to ensure the sample is representative of the full dataset), more consistent with the fact that the training set is also shuffled by default, and I believe we already released the current implementation so changing it now would be another breaking change.

This would require a different implementation though.

gleibovich-nvidia:
> thanks for working on this!
> 
> make sure to also update the CHANGELOG

This issue started with commit #73, which is after v.0.1.0 (https://github.com/NVIDIA/NeMo-Aligner/compare/v0.1.0...main). So, there's "no bug-fix" to document comparing to v0.1.0, correct?

nemo_aligner/data/nlp/builders.py
gleibovich-nvidia: Currently with `main`, PPO runs validation on a single and different batch on every validation run (as we haven't come around to fixing #27 yet). It makes no sense. 

While I also agree we should shuffle the validation set by default, I do think it's better keep it for a different issue/PR together with the accompanying expected behavior/implementation discussion, and maybe do that after we also fix #27. I wouldn't want to delay this issue being fixed, as currently validation is broken for PPO.

nemo_aligner/data/nlp/builders.py
gleibovich-nvidia: Sure. Fixed.

examples/nlp/gpt/train_gpt_ppo_actor.py
gleibovich-nvidia: Done

gleibovich-nvidia: Done

nemo_aligner/data/nlp/builders.py
odelalleau: Ah, I may be missing something here. Since `limit_val_batches` isn't supported in PPO I thought we'd always run validation on the full validation set and thus it shouldn't actually matter... what's actually happening?

(on my side I've just been using a small -- single-batch -- validation set so I didn't run into this issue)

odelalleau:
> > thanks for working on this!
> > make sure to also update the CHANGELOG
> 
> This issue started with commit #73, which is after v.0.1.0 ([v0.1.0...main](https://github.com/NVIDIA/NeMo-Aligner/compare/v0.1.0...main)). So, there's "no bug-fix" to document comparing to v0.1.0, correct?

It looks like we need to do some synch / clean-up on versions. I just checked the 24.01 NeMo framework container and we released a v0.2.0 version (according to the CHANGELOG in the container) but this hasn't been reflected in `main`. Also the container version didn't update `nemo_aligner/package_info.py` which may add to the confusion.

@gshennvm are you aware of these issues?

gshennvm:
> > > thanks for working on this!
> > > make sure to also update the CHANGELOG
> > 
> > 
> > This issue started with commit #73, which is after v.0.1.0 ([v0.1.0...main](https://github.com/NVIDIA/NeMo-Aligner/compare/v0.1.0...main)). So, there's "no bug-fix" to document comparing to v0.1.0, correct?
> 
> It looks like we need to do some synch / clean-up on versions. I just checked the 24.01 NeMo framework container and we released a v0.2.0 version (according to the CHANGELOG in the container) but this hasn't been reflected in `main`. Also the container version didn't update `nemo_aligner/package_info.py` which may add to the confusion.
> 
> @gshennvm are you aware of these issues?

yes, thanks for noticing it. I'm working on that -- need to find good version bumps for our own dockerfile

nemo_aligner/data/nlp/builders.py
odelalleau: > what's actually happening?

Ah, I just talked to @gshennvm who reminded me of `num_val_samples` which I had forgotten about. I understand what's going on now.

I'm good with doing this (the shuffling of the validation set) in a subsequent PR then. @gshennvm and I agreed on how to implement it and I should be able to take care of it next week.

odelalleau (approved):
I think we can merge the PR to unblock the PPO issue, and update the CHANGELOG later after the versioning confusion is addressed (I'm taking a note of it to be sure I don't forget).

gleibovich-nvidia:
> I think we can merge the PR to unblock the PPO issue, and update the CHANGELOG later after the versioning confusion is addressed (I'm taking a note of it to be sure I don't forget).

Thanks for looking into this! 
I think we're ready to merge now, but it seems @gshennvm has to approve first (on the requested change for updating the CHANGELOG).

gshennvm (approved):
thanks for the change! looksg ood to me

PR #116, opened 2024-02-29
Dev

gshennvm:
# What does this PR do ?

upgrade dockerfile and package versions to latest nemo release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) 
 ## [Next Version]
 
 ### New features and optimizations
+
+### Breaking changes
+
+### Bug Fixes
+
+## [0.2.0] - 2024-02
+### New features and optimizations
 - Added public-facing official Dockerfile for NeMo-Aligner.
 - PPO: memory optimization to help avoid OOM in the actor when sending training data to the critic.
 - PPO: it is now possible to use a custom end string in `sampling_params.end_strings` that is different from `<extra_id_1>`.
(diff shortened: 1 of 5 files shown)


nemo_aligner/package_info.py
```diff
@@ -14,9 +14,9 @@
 
 
 MAJOR = 0
-MINOR = 1
+MINOR = 2
 PATCH = 0
-PRE_RELEASE = "rc0"
+PRE_RELEASE = ""
```
odelalleau, line 19: Should we use something like `0.3.0.dev` since technically we are now working on the next version in `main`?

odelalleau (approved):
Looks good to me, I assume the new Dockerfile matches the one from the new NeMo.

nemo_aligner/package_info.py
gshennvm: yup! fixed https://github.com/NVIDIA/NeMo-Aligner/commit/62005260051edb5d9ea4794fa1fc3b170155e239

CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Next Version]
 
+### New features and optimizations
+
+### Breaking changes
+
+### Bug Fixes
+- Fixed issue where random sampler keeps state when resetting for validation, leading to a different validation batch each validation step. Fixed by using a deterministic sampler
```
gshennvm, line 13: @odelalleau also fixed this for comment with the random sampler fixes

odelalleau: Ok for me -- I may rewrite it when I implement the version we discussed.

odelalleau (approved):

PR #120, opened 2024-03-06
fix dpo val check interval

gshennvm:
# What does this PR do ?

fix dpo val check interval

### DIFF
--- a/nemo_aligner/algorithms/dpo.py
+++ b/nemo_aligner/algorithms/dpo.py
@@ -104,7 +104,7 @@ class DPOTrainer:
 
         self.limit_val_batches = compute_limit_batches(len(val_dataloader), self.cfg.limit_val_batches)
         self.val_check_interval = (
-            int(self.cfg.val_check_interval * self._train_dataloader_len)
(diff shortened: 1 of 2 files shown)


nemo_aligner/algorithms/dpo.py
```diff
@@ -104,7 +104,7 @@ def __init__(
 
         self.limit_val_batches = compute_limit_batches(len(val_dataloader), self.cfg.limit_val_batches)
         self.val_check_interval = (
-            int(self.cfg.val_check_interval * self._train_dataloader_len)
+            int(self.cfg.val_check_interval * len(train_dataloader))
```
trias702, line 107: Sorry to be a pain, but can we please change this to `len(self.train_dataloader)` ?

trias702 (changes requested):

nemo_aligner/algorithms/dpo.py
gshennvm: 01fcea56dad3113c06aada3c608cd63d124e8f16

PR #124, opened 2024-03-13
fix crash when val check interval is float in dpo config

gshennvm:
# What does this PR do ?

fix crash when val check interval in DPO is a float


trias702 (approved):
LGTM, sorry I missed this before

gshennvm:
> LGTM, sorry I missed this before

np! thanks for checking this

PR #135, opened 2024-03-26
Change default value of `drop_last` to True for SFT validation set

odelalleau:
# What does this PR do ?

Change default value of `drop_last` to True for SFT validation set (since False is not supported anyway and systematically crashes).

# Changelog 
- No changelog update since nobody could run `drop_last=False` anyway


gshennvm (approved):

gshennvm:
thanks for the change! you can merge once DCO issue is resolved

Issue #46, opened 2023-12-05
Log probs in PPO do not account for custom `sampling_params` 


odelalleau:
**Describe the bug**

If non-default sampling params are used, this will modify rollout generation, but the logprobs of pi(generated_text) will still be computed as if the default sampling params had been used. This introduces a mismatch between the policy being optimized and the data we generate from it, which could hurt PPO optimization.

**Expected behavior**

`sampling_params` should be taken into acount when computing log probs.

Note that we might also want to allow specifying `sampling_params` for the reference policy: for instance, if we want to sample with a different temperature, it could make sense to consider that the reference policy should use a similar temperature. In general though we don't want to enforce using the same `sampling_params` for the reference policy as it could lead to infinite KL (e.g. when using `top_k`).

PR #136, opened 2024-03-26
fixing issue 46

guyknvda:
# What does this PR do ?
Apply sampling params to the logprobs of the response tokens (see issue #46)

The application of sampling params is done by default.
to be consistent with the response generation process (done in [text_generation_utils.py](https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/nlp/modules/common/text_generation_utils.py#L737) )
the following parameters were taken into account:
- temperature
- top_p
- top_k

note that:
1. if `use_greedy` is set to True (default), the generation doesnt change the logits, thus the original logits are used to compute the log prob, ignoring the other sampling params (top_p, temperature and top_k)
2. the repetition_penalty is currently not taken into account since during the generation, it is also not taken into account (potential issue. its only taken into account if `compute_logprob` is True). 

# Additional Information
* Related to #46

### DIFF
--- a/nemo_aligner/models/nlp/gpt/megatron_gpt_ppo_actor.py
+++ b/nemo_aligner/models/nlp/gpt/megatron_gpt_ppo_actor.py
@@ -29,6 +29,12 @@ from nemo.collections.nlp.modules.common.megatron.utils import (
     get_iterator_k_split,
     get_ltor_masks_and_position_ids,
 )
+
+from nemo.collections.nlp.modules.common.text_generation_utils import (
+    repetition_penalty,
+    top_k_logits,
+)
+
 from nemo.collections.nlp.parts.utils_funcs import get_last_rank
 from nemo_aligner.models.alignable_interface import AlignableGenerativeInterface
 from nemo_aligner.utils.distributed import (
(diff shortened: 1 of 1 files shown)


nemo_aligner/models/nlp/gpt/megatron_gpt_ppo_actor.py
```diff
@@ -200,14 +201,35 @@ def finish_training(self):
         """no need to offload adam states here
         """
 
+    def _apply_sampling_params(self, logits, context_lengths):
+        samparams = self._sampling_params
+        # apply the sampling params to the logits - focusing only on the generated tokens.
+        context_length = context_lengths.min().item()
+        resp_logits = logits[:, context_length - 1 :].contiguous()
+        if not samparams.get("use_greedy", False):  # if use_greedy is True, use the logits as is
```
odelalleau, line 209: Minor suggestion: move this up two lines and write it
```python
if samparams.get("use_greedy", False):
    return logits
```
which will avoid a couple of useless ops & extra indent

odelalleau: Also: I think we should add a check to skip scaling if temp == 1, and top_p / top_k if they are equal to 1.0 / 1. This way we don't mess with logits for no good reason.

odelalleau (changes requested):
Sincere apologies for the late review!

I really appreciate you tackling this issue, which is not trivial. Unfortunately it's more complex than that, for two reasons:

1. We need to handle tensor parallelism, which means that modifying the logits probably needs to be done within `DistributedLogprob`. It's likely be going to be a bit tricky to implement but it should be doable, at the expense of a few more steps to handle top_k / top_p. Note that it may be more efficient (and less memory intensive) to gather only the top_k logits from each rank.
2. We need to also modify the logits used in the loss [here](https://github.com/NVIDIA/NeMo-Aligner/blob/003b655fd90f647c4ec06c2b3a3ab8c51d3234bc/nemo_aligner/models/nlp/gpt/megatron_gpt_ppo_actor.py#L108)

I also think we should add flags to control where exactly these transformations are applied. I'm actually not sure it's a good idea to apply it to compute the KL penalty term because:
* If we apply it to the reference policy, it may lead to infinite KL due to top_p / top_k (when we sample a token that has zero probability under the reference policy)
* If we don't apply it to the reference policy, then we may start with a high KL penalty from the start, which could cause some issues.

I would thus suggest to add some fine-grained control on where we apply this transformation, with the following default values:
```yaml
model:
  ppo:
    transform_logits_from_sampling_params:
      loss: True
      kl_penalty_actor: False
      kl_penalty_ref: ${.kl_penalty_actor}
```
This way we will be able to easily experiment with various configurations to see what actually works best in practice.

guyknvda:
replaced by new PR #186

PR #139, opened 2024-04-02
add moe support in reward model

gshennvm:
# What does this PR do ?

Add moe support for reward model(this allows mixtral to be used). Note that by default the other models inherit MegatronGPTModel which has MoE support already. We need to add this because we use a modified GPTModel for the RM

### DIFF
--- a/nemo_aligner/models/nlp/gpt/gpt_reward_model.py
+++ b/nemo_aligner/models/nlp/gpt/gpt_reward_model.py
@@ -162,6 +162,7 @@ class GPTRewardModel(GPTModel):
         share_embeddings_and_output_weights: bool = False,
         position_embedding_type: Literal["learned_absolute", "rope"] = "learned_absolute",
         rotary_percent: float = 1.0,
+        rotary_base: int = 10000,
         seq_len_interpolation_factor: Optional[float] = None,
         output_sequence: bool = False,
         use_avg_pool: bool = False,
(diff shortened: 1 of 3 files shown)


nemo_aligner/models/nlp/gpt/gpt_reward_model.py
```diff
@@ -162,6 +162,7 @@ def __init__(
         share_embeddings_and_output_weights: bool = False,
         position_embedding_type: Literal["learned_absolute", "rope"] = "learned_absolute",
         rotary_percent: float = 1.0,
+        rotary_base: int = 10000,
```
odelalleau, line 165: Shouldn't this param also be added to the `super().__init__()` call below?

odelalleau (changes requested):
Looks good except for a small oversight I believe.

(side note: looks like you can also remove the import of `get_gpt_layer_with_transformer_engine_spec`)

nemo_aligner/models/nlp/gpt/gpt_reward_model.py
gshennvm: ah! thanks for catching it 39904668357a9559e7ca97e6980abee2ab6c12d4

gshennvm:
> (side note: looks like you can also remove the import of `get_gpt_layer_with_transformer_engine_spec`)

right: eaa763acfd1203ccbff7d61425ce5b1b92d74265

odelalleau (approved):
lgtm, thanks!

PR #146, opened 2024-04-09
Fix crash when generating from an SFT model with PP>2

### LINKED ISSUE #145: `GPTSFTModel.generate()` crashes with PP>2
**Describe the bug**

The issue is that `super().generate()` will return `None` for PP ranks from 1 to PP-1. This makes [this code](https://github.com/NVIDIA/NeMo-Aligner/blob/1e4f4cf20cbca194be60a670ebf7ab3854db2fc4/nemo_aligner/models/nlp/gpt/gpt_sft_model.py#L173) crash because `output` is `None`.

odelalleau:
Fixes #145

# What does this PR do ?

Fix crash when generating from an SFT model with PP>2

(tested with PP4)


gshennvm (approved):
LGTM, thanks!

PR #148, opened 2024-04-09
Allow computing logprobs of SFT model on input strings

odelalleau:
# What does this PR do ?

This fixes an issue where our new `generate()` method for SFT models didn't support string inputs at all. But when `compute_logprob=True`, we don't actually care about the padding & extra `predictions` key added by this method, so we can support string inputs easily by simply re-using the parent class' `generate()`.

This is useful e.g. for benchmarking purpose (this is how I tested it -- with the MMLU benchmark).

### DIFF
--- a/nemo_aligner/models/nlp/gpt/gpt_sft_model.py
+++ b/nemo_aligner/models/nlp/gpt/gpt_sft_model.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Optional, Tuple
+from typing import List, Optional, Tuple, Union
 
 import hydra
 import torch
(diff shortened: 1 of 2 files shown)


nemo_aligner/models/nlp/gpt/gpt_sft_model.py
```diff
@@ -144,17 +144,31 @@ def finish_validation_step(self):
...
+
+        1. Apply padding to max length.
+        2. Add a "predictions" key to the output, which is the model output without the prompt.
 
-        1. Going to apply padding to max length
-        2. Going to append "predictions" key which is the model output without the prompt
+        These two additional steps above are only performed for actual generation from the model:
+        if `generate()` is called with `compute_logprob=True` then the base model method is used.
         """
+        if sampling_params is not None and sampling_params.get("compute_logprob", False):
```
gshennvm, line 162: is compute_logprob=True mutually exclusive with generating new tokens? maybe we should guard the inputs being string in this check too

odelalleau: Yeah so as soon as you set `compute_logprob=True` it automatically set other values, like generating a single token: https://github.com/NVIDIA/NeMo/blob/d8ebaa5fb3b7df556b282f3547a605bded5b61de/nemo/collections/nlp/modules/common/text_generation_utils.py#L101

It's supposed to work with tensors as input too (not just strings).

gshennvm: ah okay! that makes sense

gshennvm (approved):

PR #149, opened 2024-04-10
trtllm0.9 changes

jiemingz:
# What does this PR do ?

# Changelog 

# Usage

```python
```

**Pre checks**:

# Additional Information
* Related to # (issue)

### DIFF
--- a/examples/nlp/gpt/conf/gpt_ppo_actor.yaml
+++ b/examples/nlp/gpt/conf/gpt_ppo_actor.yaml
@@ -35,14 +35,18 @@ trainer:
     # but will not crash and training will still proceed
     rollout_batch_seq_length: null
 
+    # Accelerate training times by accelerating inference stage using TRTLLM
     trt_llm:
       enable: False
-      reshard: False # if True then reshard the model into TP only for inf
+      reshard: True # if True then reshard the model into TP only for inference
       max_context_len: ${int_div:${model.encoder_seq_length}, 2}
-      model_type: LLaMAForCausalLM
 
-      # Unload the TRTLLM engine during the training stage 
-      # (TODO: Current not supported)
+      # The model type TRTLLM will build, supported models are listed at:
+      # https://github.com/NVIDIA/TensorRT-LLM/blob/v0.9.0/tensorrt_llm/models/__init__.py#L75
+      model_type: LLaMAForCausalLM 
+
+      # Unload and reload the TRTLLM engine before and after the training stage 
+      # Reloading the engine incurs a constant time overhead
       unload_engine_train: False
 
     flask_server:
--- a/examples/nlp/gpt/conf/gpt_ppo_actor.yaml
+++ b/examples/nlp/gpt/conf/gpt_ppo_actor.yaml
@@ -38,6 +38,12 @@ trainer:
     trt_llm:
       enable: False
       reshard: False # if True then reshard the model into TP only for inf
+      max_context_len: ${int_div:${model.encoder_seq_length}, 2}
+      model_type: "LLaMAForCausalLM"
+
+      # Unload the TRTLLM engine during the training stage 
+      # (TODO: Current not supported)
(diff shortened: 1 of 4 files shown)


nemo_aligner/utils/trt_llm.py
```diff
@@ -1,136 +1,111 @@
...
+        output_ids = output_dict['output_ids']
+        resp_lens = output_dict['sequence_lengths'].squeeze()
 
-        prompt_tokens, prompt_lengths = inputs
+        # remove beam dim from output_ids: [mbs, beam_dim, sequence len]
+        mbs = output_ids.shape[0]
+        if mbs == 1:
+            output_ids = output_ids.view([1,output_ids.shape[-1]])
+        else:
+            output_ids = output_ids.squeeze()
```
gshennvm, line 85: can't you do `torch.atleast_2d(output_ids).long()` here?

nemo_aligner/utils/trt_llm.py
```diff
@@ -1,136 +1,111 @@
...
+        output_ids = output_ids.to(torch.int64)
 
-        batch_input_ids = []
-        for idx in range(prompt_tokens.shape[0]):
-            batch_input_ids.append(prompt_tokens[idx][0 : prompt_lengths[idx]].cpu())
+        #broadcast output to all PP ranks
+        if parallel_state.get_pipeline_model_parallel_world_size() > 1 and not self.reshard_model:  
+            group = parallel_state.get_pipeline_model_parallel_group()
+            src = parallel_state.get_pipeline_model_parallel_first_rank()
+            output_ids = broadcast_2d_tensor(output_ids, src, group, dtype=output_ids.dtype)
```
gshennvm, line 92: https://github.com/NVIDIA/NeMo-Aligner/blob/geshen/main_trt/nemo_aligner/utils/distributed.py#L102

I think you can use this function instead. The groups are unchanged when we don't have resharding and when there's resharding this should be fairly trivial(though if you think we have perf hit lmk)

examples/nlp/gpt/conf/gpt_ppo_actor.yaml
```diff
@@ -38,6 +38,12 @@ trainer:
     trt_llm:
       enable: False
       reshard: False # if True then reshard the model into TP only for inf
+      max_context_len: ${int_div:${model.encoder_seq_length}, 2}
+      model_type: "LLaMAForCausalLM"
```
gshennvm, line 42: oh... forgot about this? we need to change this for mixtral etc?

nemo_aligner/utils/trt_llm.py
jiemingz: wouldnt this also broadcast to gpus in the same TP group? Which TRTLLM already has equal across TP

gshennvm: yeah do you think that's too much of a perf hit or doesn't matter?

JimmyZhang12: I dont think it would affect perf but it does make the code less clear
This broadcast only happens without resharding, do we need to guard it?

nemo_aligner/utils/trt_llm.py
JimmyZhang12: From pytorch doc it says ` Input tensors with two or more dimensions are returned as-is.` and output_ids is 3d

nemo_aligner/utils/trt_llm.py
gshennvm: okay let's guard on no reshard and then use that function then?

examples/nlp/gpt/conf/gpt_ppo_actor.yaml
gshennvm: okay it'll be good to just list which ones people can specify

gshennvm (approved):
LGTM once you resolve the 2 small comments

PR #152, opened 2024-04-12
Use `top_p=0.0` by default (instead of 1.0)

odelalleau:
This is because we otherwise do useless computations, at least until
        https://github.com/NVIDIA/NeMo/pull/8905
is merged.


gshennvm:
hm, I'm a bit hesitant to merge this because we're going to forget to revert it back again. Should we wait for your nemo PR to be merged and then cherrypick it into our dockerfile?

odelalleau:
> we're going to forget to revert it back again

We don't need to. `0.0` is fine and what we should have been using from the start (I guess we just forgot to check how it was implemented in NeMo).

gshennvm:
> > we're going to forget to revert it back again
> 
> We don't need to. `0.0` is fine and what we should have been using from the start (I guess we just forgot to check how it was implemented in NeMo).

ah I see. then yes this LGTM. it's less confusing this way

gshennvm:
could you also update the CHANGELOG to describe the issue?

PR #154, opened 2024-04-15
Constitutional AI: Harmlessness from AI Feedback

gleibovich-nvidia:
Draft for the implementation of  [Constitutional AI: Harmlessness from AI Feedback](https://arxiv.org/abs/2212.08073). 

Note that there are still some minor missing pieces regarding the exact model and datasets to be used, yet the main building blocks are in place.


examples/nlp/cai/generate_rl_cai_dataset.py
```diff
@@ -0,0 +1,741 @@
...
+        "temperature": temperature,
+        "add_BOS": True,
+        "top_k": 50,
+        "top_p": 0.95,
+        "greedy": False,
+        "all_probs": False,
+        "repetition_penalty": 1,
+        "min_tokens_to_generate": 1,
+        "end_strings": ["<extra_id_1>"],
+    }
```
terrykong, line 242: Can you please parametrize this json obj? If someone wants to try generating 2048 tokens, it would be helpful to not have to edit code. 

hydra may be too heavy for this script, so perhaps plumbing these via [add_argument_group](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_argument_group) would be a good middle ground.

examples/nlp/cai/generate_rl_cai_dataset.py
```diff
@@ -0,0 +1,741 @@
...
+    headers = {
+        "Authorization": f"Bearer {api_key}",
+        "Accept": "application/json",
+    }
+
+    payload = {
+        "messages": messages,
+        "temperature": temperature + 0.0000001,
+        "top_p": 0.0000001,
+        "max_tokens": 1024,
```
terrykong, line 386: Same comment about parametrizing these arguments

examples/nlp/cai/generate_rl_cai_dataset.py
```diff
@@ -0,0 +1,741 @@
...
+class ChatPromptTemplate:
+    system_token = "System"
+    user_token = "User"
+    assistant_token = "Assistant"
+
+    system_turn_token = "<extra_id_0>"
+    turn_token = "<extra_id_1>"  # (model_config: chat_prompt_tokens.turn_start)
+    end_signal = "\n"  # (model_config: chat_prompt_tokens.end_of_turn. NOTE: "\x0A" is '\n' in ASCII code)
+    label_start = "<extra_id_2>"  # (model_config: chat_prompt_tokens.label_start)
+    end_name_signal = "\n"  # (model_config: chat_prompt_tokens.end_of_name. NOTE: "\x0A" is '\n' in ASCII code)
```
terrykong, line 25: Can we parametrize these and reuse the variable holding them throughout this script. `extra_id_*` is hardcoded in many places and it would be cleaner to reference a variable. WDYT about plumbing it from the parser?

examples/nlp/cai/generate_sl_cai_dataset.py
```diff
@@ -0,0 +1,484 @@
...
+        return resp_sentences
+
+    data = {
+        "sentences": prompt,
+        "add_BOS": True,
+        "top_k": 1,
+        "top_p": 0.9,
+        "all_probs": False,
+        "repetition_penalty": 1.2,
+        "min_tokens_to_generate": 1,
```
terrykong, line 87: WDYT about parametrizing this?

examples/nlp/cai/process_anthropic_hh_using_chat_prompt.py
```diff
@@ -0,0 +1,243 @@
...
+            else:
+                prompt_string_to_use = prompt_continuation_format.format(text=string_to_use, body=body, response="")
+                string_to_use = prompt_continuation_format.format(text=string_to_use, body=body, response=response)
+
+        # just in case... make sure we have single backslash
+        prompt_string_to_use = prompt_string_to_use.rstrip("\n") + "\n"
+        string_to_use = string_to_use.rstrip("\n") + "\n"
+
+        if add_eos:
+            string_to_use += "<extra_id_1>"
```
terrykong, line 120: Can this be parametrized?

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,298 @@
...
+   
+   .. code-block:: bash
+   
+        python examples/nlp/cai/process_anthropic_hh_using_chat_prompt.py
+           --output-dir /path/to/anthropic_helpful_only
+           --dataset-dir-name helpful-base helpful-online helpful-rejection-sampled
+           --output-file-name-prefix anthropic_helpful_only
+
+
+Step 2: Generate and revise responses to harmfulness prompts creating the SL-CAI dataset
```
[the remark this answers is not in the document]
snisimov: done

examples/nlp/cai/generate_rl_cai_dataset.py
```diff
@@ -0,0 +1,741 @@
...
+            samples.extend(rlaif_batch_samples)
+
+            samples_per_temperature[str(t)] = samples
+
+        all_samples.extend(join_responses(samples_per_temperature))
+
+    return all_samples
+
+
+def generate_responses_batch(prompt_list: list, temperature: int, port_num: int):
```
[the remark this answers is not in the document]
snisimov: fixed

examples/nlp/cai/generate_sl_cai_dataset.py
```diff
@@ -0,0 +1,484 @@
...
+
+    args = parser.parse_args()
+    assert os.path.isfile(args.red_teaming_prompts_dataset_path)
+    assert os.path.isfile(args.few_shot_prompts_dataset_path)
+    assert os.path.isfile(args.helpfulness_dataset_path)
+    assert args.num_examples >= -1 and args.num_examples != 0
+
+    if args.max_seq_length is not None:
+        assert args.max_seq_length > 0
+        assert os.path.isfile(args.tokenizer_model)
```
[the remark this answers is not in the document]
snisimov: done

examples/nlp/cai/generate_rl_cai_dataset.py
```diff
@@ -0,0 +1,741 @@
...
+
+
+def model_remote_inference(prompt, port_num=5656, temperature=1.0):
+    if not isinstance(prompt, list):
+        prompt = [prompt]
+
+    headers = {"Content-Type": "application/json"}
+
+    def request_data(request):
+        resp = requests.put("http://localhost:{}/generate".format(port_num), data=json.dumps(request), headers=headers)
```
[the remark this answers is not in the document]
snisimov: The service needs to be started by the user. I've added an argument to the parser to specify the hostname or IP address of the inference service, which allows the user to define where the service is running.

examples/nlp/cai/generate_rl_cai_dataset.py
```diff
@@ -0,0 +1,741 @@
...
+        return prompt_with_response
+
+    @staticmethod
+    def extract_response(prompt: str):
+        response = prompt.rsplit(ChatPromptTemplate.assistant_message_header, 1)[-1]
+        response = response.strip().removesuffix(ChatPromptTemplate.turn_token).strip()
+        return response
+
+
+constitution = """
```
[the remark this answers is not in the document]
gleibovich-nvidia: Good point that we definitely need to discuss. 
As CAI paper is 1.5 years old by now [= forever in LLM developement :-)] and has no official code implementation, we agreed with @gshennvm that we will roughly follow the paper and might make some changes where we see fit. Given that the supported context size for recent models is getting increasingly larger each day, we experimented with feeding the entire constitution to the judge LLM and have it judge according to all the principles at once. Empirically we found that working quite well, as the selected most harmless and most harmful responses made perfect sense on several tens of prompts that we experimented with. 

What do you think? @terrykong @gshennvm 

P.S: there are other section which is not exactly following the paper, and we might want to change those as well if we want to follow the paper exactly.  e.g. in the paper in order to rank and choose between the responses, they use the logprob of each of the responses as the criteria for selection, whereas in our implementation we use an external judge LLM as, for instance, is being done with MT-Bench for scoring LLMs.

examples/nlp/cai/generate_sl_cai_dataset.py
```diff
@@ -0,0 +1,484 @@
...
+    assert port_num >= 0
+    assert prompt is not None and isinstance(prompt, (str, list))
+    if not isinstance(prompt, list):
+        prompt = [prompt]
+    assert all(isinstance(p, str) for p in prompt)
+
+    headers = {"Content-Type": "application/json"}
+
+    def request_data(request):
+        resp = requests.put("http://localhost:{}/generate".format(port_num), data=json.dumps(request), headers=headers)
```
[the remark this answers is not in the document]
snisimov: The service needs to be started by the user. I've added an argument to the parser to specify the hostname or IP address of the inference service, which allows the user to define where the service is running.

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,298 @@
...
+###################################################################################################
+
+Run an inference server in the background using the following command:
+
+.. code-block:: bash
+
+   python /opt/NeMo/examples/nlp/language_modeling/megatron_gpt_eval.py
```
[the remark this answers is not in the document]
gleibovich-nvidia: done

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,298 @@
...
+           trainer.devices=8 
+           trainer.num_nodes=1 
+           web_server=False 
+           port=5999 
+
+Please wait for the server to be ready before proceeding.
+
+.. code-block:: bash
+
+   python examples/nlp/cai/generate_sl_cai_dataset.py
```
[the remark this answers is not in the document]
gleibovich-nvidia: done

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,298 @@
...
+
+
+Step 3: Finetune Mistral-7B on the revised responses to create a Mistral-7B-SL-CAI model
+######################################################################################################
+
+Note that you would need to set up multi-node training run in your cluster env, depending on the type of cluster you use. For details, please refer to https://lightning.ai/docs/pytorch/stable/clouds/cluster.html
+
+.. code-block:: bash
+
+   python examples/nlp/gpt/train_gpt_sft.py
```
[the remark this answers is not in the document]
gleibovich-nvidia: done

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,298 @@
...
+
+
+Step 4: Generate the RL-CAI (preference) dataset for RM and PPO training
+##############################################################################################################
+
+Run an inference server in the background using the following command:
+
+.. code-block:: bash
+
+   python /opt/NeMo/examples/nlp/language_modeling/megatron_gpt_eval.py
```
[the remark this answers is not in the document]
gleibovich-nvidia: done

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,298 @@
...
+           trainer.num_nodes=1
+           web_server=False
+           port=5999
+
+Please wait for the server to be ready before proceeding.
+
+
+.. code-block:: bash
+
+   python examples/nlp/cai/generate_rl_cai_dataset.py
```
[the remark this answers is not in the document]
gleibovich-nvidia: done

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,298 @@
...
+
+(Supervised Stage) Critique → Revision → Supervised Learning: The AI generates responses to harmfulness prompts using a helpful-only AI assistant, then critiques and revises its own responses according to a principle in the constitution, and then finetunes the original model on the revised responses.
+
+(RL Stage) AI Comparison Evaluations → Reward Model → Reinforcement Learning: The AI generates pairs of responses to harmfulness prompts using the finetuned model, then evaluates which response is better according to a principle in the constitution, and then trains a reward model from this dataset of AI preferences. The AI then trains with RL using the learned reward model.
+
+.. image:: ../assets/cai_diagram.png
+   :alt: basic steps of the CAI process
+
+Both the critiques and the AI feedback are steered by a small set of principles drawn from a ‘constitution’. The supervised stage significantly improves the initial model, and gives some control over the initial behavior at the start of the RL phase, addressing potential exploration problems. The RL stage significantly improves performance and reliability.
```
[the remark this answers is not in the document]
gleibovich-nvidia: done

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,298 @@
...
+
+.. _Constitutional AI (CAI): https://arxiv.org/abs/2212.08073
+
+CAI
+###############
+The basic steps of CAI, as illustrated in Figure 1 of the paper (below):
+
+(Supervised Stage) Critique → Revision → Supervised Learning: The AI generates responses to harmfulness prompts using a helpful-only AI assistant, then critiques and revises its own responses according to a principle in the constitution, and then finetunes the original model on the revised responses.
+
+(RL Stage) AI Comparison Evaluations → Reward Model → Reinforcement Learning: The AI generates pairs of responses to harmfulness prompts using the finetuned model, then evaluates which response is better according to a principle in the constitution, and then trains a reward model from this dataset of AI preferences. The AI then trains with RL using the learned reward model.
```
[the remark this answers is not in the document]
gleibovich-nvidia: thanks, fixed

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,298 @@
...
+
+This will create the ``rl-cai`` dataset files in the defined output folder with the given output filename prefix.
+
+
+Step 5: Train the RM
+#####################
+
+.. code-block:: bash
+
+   python examples/nlp/gpt/train_reward_model.py
```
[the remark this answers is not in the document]
gleibovich-nvidia: thanks, done

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,298 @@
...
+      pretrained_checkpoint.restore_from_path='/lustre/fsw/coreai_dlalgo_llm/nemo_aligner_cai/cai_tutorial/cai_mistral_7b/3_sl_cai_model/mistral_7b_sl_cai_model.nemo'
```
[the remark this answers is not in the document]
gleibovich-nvidia: thanks! fixed.

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,298 @@
...
+      ++model.micro_batch_size=1
+      ++model.global_batch_size=64
+      ++model.seed=1234
+      exp_manager.explicit_log_dir=<path to critic output dir>
+
+Run the following command to launch actor training and reference policy server:
+
+.. code-block:: bash
+
+   python -u examples/nlp/gpt/train_gpt_ppo_actor.py
```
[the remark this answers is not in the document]
gleibovich-nvidia: thanks, done

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,298 @@
...
+           port=5999
+
+Please wait for the server to be ready before proceeding.
+
+
+.. code-block:: bash
+
+   python examples/nlp/cai/generate_rl_cai_dataset.py
+      --batch-size 128
+      --ngc-api-key nvapi-**_YOUR-NGC-KEY-GOES-IN-HERE_**
```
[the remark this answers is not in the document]
gleibovich-nvidia: Yes. Thanks, added it.

examples/nlp/cai/utils.py
```diff
@@ -0,0 +1,89 @@
+import os
+from collections import defaultdict
+from multiprocessing import Pool
+
+from tqdm import tqdm
```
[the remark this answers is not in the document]
gleibovich-nvidia: NeMo-Aligner is run from within the NeMo container, and 'tqdm' is present in the NeMo container's [requirements.txt](https://github.com/NVIDIA/NeMo/blob/8e65042d15062ce3fbe639f9d428c639510d894c/requirements/requirements.txt#L13)

examples/nlp/cai/mistral7b_few_shot_samples
```diff
@@ -0,0 +1,53 @@
+<s>[INST] Give me some ideas to prank my boss? [/INST]1. Change your email signature to something humorous or inappropriate.
```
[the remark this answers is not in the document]
gleibovich-nvidia: We found that the few shot prompting samples significantly improve the results of the SL-CAI part of the flow (as is also mentioned in the paper).

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,298 @@
...
+
+7. Inference.
+
+.. image:: ../assets/cai_flow.png
+
+Step 1: Download models and datasets
+#############################################################################
+1. Download ``Mistral-7B-Instruct`` and ``Mistral-7B`` LLM models from https://huggingface.co/mistralai/Mistral-7B-Instruct-v0.1 and https://huggingface.co/mistralai/Mistral-7B-v0.1 into the models folder.
+
+   Then convert into .nemo format:
```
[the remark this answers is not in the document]
gleibovich-nvidia: Sure, I can add it. 
Although do note that other tutorials/algorithms also do not point for where to get a container. 

e.g.
https://github.com/NVIDIA/NeMo-Aligner/blob/cai/docs/user-guide/RLHF.rst?plain=1#L105
https://github.com/NVIDIA/NeMo-Aligner/blob/cai/docs/user-guide/SteerLM.rst?plain=1#L58

I'm guessing that the assumption might be that NeMo-Aligner is already run from within a NeMo container, and thus no need to point the user to where to get it from. 

@gshennvm What do you think? Is there a reason we do not specifically point the user to where to get a container from?

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,298 @@
...
+      pretrained_checkpoint.restore_from_path='/lustre/fsw/coreai_dlalgo_llm/nemo_aligner_cai/cai_tutorial/cai_mistral_7b/3_sl_cai_model/mistral_7b_sl_cai_model.nemo'
+      "++model.data.data_prefix={train: [<path to cai_preference_dataset_comparisons_train.jsonl>], validation: [<path to cai_preference_dataset_comparisons_test.jsonl>], test: [<path to cai_preference_dataset_comparisons_test.jsonl>]}"
+      ++model.micro_batch_size=1
+      ++model.global_batch_size=512
+      ++model.activations_checkpoint_granularity="full"
+      ++model.activations_checkpoint_method="uniform"
+      ++model.activations_checkpoint_num_layers=1
+      ++model.optim.bucket_cap_mb=200
+      ++model.data.seq_length=4096
+      ++model.data.data_impl=json
```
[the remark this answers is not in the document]
gleibovich-nvidia: thanks, fixed

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,298 @@
...
+      exp_manager.explicit_log_dir=<path to actor output dir>
+      ++model.micro_batch_size=1
+      ++model.global_batch_size=64
+      ++model.activations_checkpoint_granularity=selective
+      ++model.activations_checkpoint_method=uniform
+      ++model.optim.lr=9e-8
+      ++model.optim.sched.min_lr=9e-9
+      ++model.ppo.num_rollout_samples=512
+      ++model.ppo.rollout_micro_batch_size=8
+      ++model.ppo.length_params.max_length=1024
```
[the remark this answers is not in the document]
gleibovich-nvidia: thanks, fixed

examples/nlp/cai/generate_rl_cai_dataset.py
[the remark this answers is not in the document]
terrykong: I personally think it's fine to deviate from the algorithm as long as it's documented; at the very least in a comment.

I do think the constitutions should be passed into the script as opposed to being a literal within the python script. It makes customization more difficult since we are mixing "data" with code

examples/nlp/cai/utils.py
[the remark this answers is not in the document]
terrykong: This is more of a nit, since I highly doubt `tdqm` will ever be removed. I just wanted to point out that it's better practice to declare the packages you explicitly import and let `pip` decide if a re-install is needed. You could simply leave it unpinned in https://github.com/NVIDIA/NeMo-Aligner/blob/main/requirements.txt. What you're guarding against is if upstream repos don't use that package anymore, then your code is broken. It also makes explicit what your project expects

examples/nlp/cai/mistral7b_few_shot_samples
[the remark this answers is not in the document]
terrykong: My intention is just to reduce the risk of scrutiny by including these particular few shot examples. The options I see are:
1. just link to the few shot examples in the constitutionalAI repo
2. rewrite the prompts to be objective and without controversy
3. Just omit the few shot examples and explain to the user the format

I think (3) is the easiest, but puts the onus on the user. @gshennvm WDYT?

FWIW, (3) also has the benefit of documenting how to even create this file, which at the moment it's not clear to me how to do it given the prompt templates you're introducing and whether the special tokens need to be present in some way.

examples/nlp/cai/generate_rl_cai_dataset.py
snisimov: Thank you for your suggestion.
I've implemented the **add_argument_group** approach to parameterize the JSON object as you recommended.

examples/nlp/cai/generate_sl_cai_dataset.py
snisimov: Thank you for your suggestion.
I've implemented the add_argument_group approach to parameterize the JSON object as you recommended.
i will do the same for 'examples/nlp/cai/generate_rl_cai_dataset.py'

examples/nlp/cai/generate_rl_cai_dataset.py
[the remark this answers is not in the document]
gleibovich-nvidia: Absolutely agree. Will fix and make sure to properly document! 
Thanks for the feedback.

examples/nlp/cai/mistral7b_few_shot_samples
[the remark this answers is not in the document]
gleibovich-nvidia: Oh sure, I get your point now. 

I think option (1) is less preferred as:
a. The few-shot prompts should be formatted with the matching prompt template of the LLM that we're going to align (Mistral-7B in this particular case), which the original CAI repo would be lacking.
b. It's probably better that the responses are in-distribution (on-policy), i.e. a result of generation by the LLM model to be aligned.

I'm fine with either (2) or (3), and agree regarding the benefit of (3) leading to better documentation on how to generate these.

examples/nlp/cai/utils.py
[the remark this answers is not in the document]
gleibovich-nvidia: While I do agree, I would like to again stress that, as with the container, if this is something that we would like to address then this is not specific to this algorithm/tutorial, and we might want to fix it framework-wide. It is not a new requirement. Thus, I'm wondering if there was some other consideration not to add it so far. 

e.g. `tqdm` is also being used by other algorithms within NeMo-Aligner and wasn't added to the `requirements.txt`. 
https://github.com/NVIDIA/NeMo-Aligner/blob/main/examples/nlp/data/steerlm/attribute_annotate.py#L36

examples/nlp/cai/generate_rl_cai_dataset.py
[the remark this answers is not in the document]
gleibovich-nvidia: Done. Constitution is now in a separate file and paper deviation is properly documented.

examples/nlp/cai/generate_rl_cai_dataset.py
```diff
@@ -0,0 +1,741 @@
...
+First, thoroughly review all responses.
+Your output should be a simple dictionary:
+{ "chosen" : x, "rejected": y}
+, where x, y are the response ids. NOTE: you must assign a single number to both ids, x and y.
+
+Please ensure your output is presented first, followed by an explanation.
+\n
+"""
+
+system_prompt = f"{prefix, constitution}"
```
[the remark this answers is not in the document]
gleibovich-nvidia: Thanks, renamed and fixed.

examples/nlp/cai/mistral7b_few_shot_samples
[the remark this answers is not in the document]
gleibovich-nvidia: I guess option (2) is a good compromise which is still run-able for the user, but I'm not entirely sure what would be the impact on the SL-CAI dataset and the trained model if the prompts and/or responses would be too neutral. We can try and rerun the whole pipeline after the change and check for the actual impact (but that might delay the release a bit). 
On the other hand, I mean, that's the whole point of this algorithm, to deal with toxic prompts/responses, so maybe it's not too bad if these examples are left as-is? 
 Anyway, I'm fine with both leaving this unchanged, or with option (2) or option (3).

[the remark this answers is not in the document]
gleibovich-nvidia: Hi @terrykong, option (3) is implemented here: 
https://github.com/NVIDIA/NeMo-Aligner/pull/154/commits/439a389d6a6fd56c73686f5916b05b1850c3e89a

What do you think? I'm fine with it, as it allows us the flexibility we need with the few-shot samples.

examples/nlp/cai/utils.py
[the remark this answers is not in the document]
terrykong: Since I classify this as a nit, let me just resolve this discussion as "fix later"; this shouldn't be a reason not to merge this.

examples/nlp/cai/utils.py
```diff
@@ -0,0 +1,514 @@
...
+        if self.system_default_message is not None and "System" in self.roles:
+            # NOTE: It is assumed that if a system message exists, it should be the first message.
+            if messages[0]["role"] != "System":
+                messages = [{"content": self.system_default_message, "role": "System"}] + messages
+
+        return super().format_message(messages)
+
+
+if __name__ == "__main__":
+    extra_id_prompt_template = PromptTemplate(
```
terrykong, line 424: What's this main method for? It looks very nearly like what you'd write for a test. If it's just for debug, can we convert this to `utils_test.py` and use pytest to validate the templating?

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+      --tokenizer-library sentencepiece \
+      --tokenizer-model /path/to/<TOKENIZER_FILE_NAME>.model \
+      --helpfulness-dataset-path /path/to/nvidia_sft_datablend_v1_train.json \
+      --output-filepath /path/to/cai_revisions_aligner_chat_template.jsonl \
+      --port 5999 \
+      --host <hostname or IP address of the inference service>
+
+This will generate an SL-CAI dataset of prompts and revised responses as ``cai_revisions_aligner_chat_template.json``
+
+The few-shot samples should be provided following the template in ``few_shot_samples_example.json``, and should include a red teaming prompt, a response from the helpful model (e.g. ``Mistral-7B`` in this tutorial), critique and revision requests and responses. The Anthropic repo outlines an example `here <https://github.com/anthropics/ConstitutionalHarmlessnessPaper/blob/main/prompts/CritiqueRevisionFewShotPrompts.json>`_.
```
terrykong, line 143: Just checking my understanding: is the user expected to only change the values corresponding to the `content`s key?

terrykong:
@gleibovich-nvidia Any updates on adding some unit tests? Let me know if you have any questions

examples/nlp/cai/utils.py
gleibovich-nvidia: Yes, exactly. The plan is move it to a test.

docs/user-guide/CAI.rst
gleibovich-nvidia: Yes, exactly, and potentially add several more samples if they wish (the template only include two samples, as an example). I will make this clearer.

examples/nlp/cai/generate_rl_cai_dataset.py
```diff
@@ -0,0 +1,741 @@
...
+            + ChatPromptTemplate.system_token
+            + ChatPromptTemplate.end_name_signal
+            + system_prompt
+            + ChatPromptTemplate.end_signal
+        )
+
+        return header
+
+    @staticmethod
+    def _apply_role_template(role: str, prompt: Optional[str] = None):
```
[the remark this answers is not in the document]
gleibovich-nvidia: moved to `utils/cai_utils` and added pytest tests for all of our use-cases.

examples/nlp/cai/utils.py
gleibovich-nvidia: Fixed

gleibovich-nvidia:
> @gleibovich-nvidia Any updates on adding some unit tests? Let me know if you have any questions

Sure, just added a module `tests/test_cai_utils.py` to test for extra utils functionality that this PR has added.

examples/nlp/cai/generate_rl_cai_dataset.py
snisimov: Thanks, done.
we have implemented the use of 'add_argument_group' to parameterize the JSON object as you recommended.

examples/nlp/cai/generate_rl_cai_dataset.py
snisimov: Thanks, done.
we have implemented the use of 'add_argument_group' to parameterize the variables.

examples/nlp/cai/process_anthropic_hh_using_chat_prompt.py
snisimov: Thanks, done.

gleibovich-nvidia:
> Hi @gleibovich-nvidia ! Thanks so much for this contribution, and expanding our supported algorithms.
> 
> Apologies in advance since my review is as large due to the large PR
> 
> The documentation is great.
> 
> I do think it'd be helpful to add tests for the scripts you added, in particular:
> 
> * `generate_rl_cai_dataset.py` -> `generate_rl_cai_dataset_test.py`
> * `generate_sl_cai_dataset.py` -> `generate_sl_cai_dataset_test.py`
> * `process_anthropic_hh_using_chat_prompt.py` -> `process_anthropic_hh_using_chat_prompt_test.py`
>   where the test adds a pytest that shows a valid input and expected output. My understanding is that these scripts also rely on a remote inference call, so in the test, it's best to just mock that out.
> 
> I also saw a few hard coded `extra_id_*` and I think those should be cleaned up so that it is configurable and plumbed from the user; especially since other scripts allow those to be configured to some degree.

Hi @terrykong, thanks so much for the review and excellent feedback! 
I think it helped increasing the PR quality quite significantly. 

At this point, I think that other than 3 additional comments requiring @gshennvm attention, we have addressed and fixed all the open issues. 

As for e2e tests for `generate_rl_cai_dataset.py`, `generate_sl_cai_dataset.py` and `process_anthropic_hh_using_chat_prompt.py`, let's discuss what exactly we want to get done here. 

As CAI is not a new distinct algorithm within NeMo-Aligner, and is more of a tutorial, I think (and also following a short chat with @gshennvm on this matter), we should focus on testing code functionality which is dependent on NeMo-Aligner's APIs. This is in order to make sure not to break in case one of those APIs introduce breaking changes. This is consistent with how other tutorials are implemented in the framework (e.g. SteerLM and DRaFT+), which do not implement e2e tests on their own. 

Reviewing those specific scripts, the only dependency is on `cai_utils` for which we have introduced `pytest` tests as part of this PR. There are no NeMo-Aligner APIs dependencies, as running SFT, RM and PPO is executed by the user (cmdline runs), following and in-between SL-CAI and RL-CAI datasets generation. 

 I would be happy to get your feedback and thoughts on how to proceed, and what we want to get done to finalize the last steps required to get this PR merged. 

Thanks! 
Gal

nemo_aligner/utils/cai_utils.py
terrykong, line 1: I feel that this should remain in examples/nlp/cai since the pattern thus far has been to leave utils similar to these in their respective examples folder:
* https://github.com/NVIDIA/NeMo-Aligner/blob/5b9c1f9aaeb08894be1c05bb5c9f673e8e1b4bf7/examples/nlp/data/steerlm/preprocess_helpsteer_data.py#L23
* https://github.com/NVIDIA/NeMo-Aligner/blob/5b9c1f9aaeb08894be1c05bb5c9f673e8e1b4bf7/examples/nlp/data/steerlm/preprocess_openassistant_data.py#L63

We can consider promoting it into the main package once another project requires these exact utils.

terrykong:
@gleibovich-nvidia @snisimov Thanks for addressing and iterating on the PR. I have also asked our writer to review the docs as these will surface in the Nemo FW Docs

CC @jgerh

nemo_aligner/utils/cai_utils.py
gleibovich-nvidia: Agree. The reason I moved it to `utils` is to allow for "easy" importing from `tests/test_cai_utils.py`. It is not importable otherwise, as `examples` is not part of the package. We also can't have `test_cai_utils.py` under `examples/nlp/cai` as pytest would ignore it.

There are thus three ways to resolve:
1. Have `cai_utils` under `utils` (as is it done now).
2. Add `__init__.py` to `examples` and `examples/nlp`. 
3. Add the path to dynamically to `sys.path`. 

I didn't like options (2) and (3) so much as option (2) enforces `examples` to be part of the package, which I'm not sure we want. Option (3) is also not ideal since messing around with `sys.path` might lead to unexpected hard to debug issues on the user side. 

Since, as I understand, option (1) is also less preferred, we can go with option (3). I have pushed this change now, PTAL and let me know what you think. If you have other ideas in mind, please let me know and I would be happy to follow up and implement.

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+.. _model-aligner-cai:
+
+Constitutional AI: Harmlessness from AI Feedback
+@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
+
+
+`Constitutional AI (CAI)`_ is an approach by Anthropic to train AI systems that are helpful, honest, and harmless, without relying on human feedback labels for harmlessness.
+CAI uses a set of natural language principles to guide AI behavior and self-improvement.
+The method consists of two stages: a supervised learning stage, where the AI critiques and revises its own responses according to the principles, and a reinforcement learning stage, where the AI learns from its own feedback based on the principles.
+CAI allows training a harmless but non-evasive AI assistant that engages with harmful queries by explaining its objections to them.
```
jgerh, line 12: harmless, but

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+
+.. _Constitutional AI (CAI): https://arxiv.org/abs/2212.08073
+
+CAI
+###############
+The basic steps of CAI, as illustrated in Figure 1 of the paper (below):
```
jgerh, line 18: The basic steps of CAI are described in this section and illustrated in the figure below.

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+
+(Supervised Stage) Critique → Revision → Supervised Learning: The AI generates responses to harmfulness prompts using a helpful-only AI assistant, then critiques and revises its own responses according to a principle in the constitution, and then finetunes the original model on the revised responses.
```
jgerh, line 20: fine-tunes

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+
+(RL Stage) AI Comparison Evaluations → Reward Model → Reinforcement Learning: The AI generates pairs of responses to harmfulness prompts using the finetuned model, then evaluates which response is better according to a principle in the constitution, and then trains a reward model based on this dataset of AI preferences and a human helpfulness preferences. The AI then trains with RL using the learned reward model.
+
+.. image:: ../assets/cai_diagram.png
+   :alt: basic steps of the CAI process
+
+Critiques, revisions and AI harmlessness feedback are steered by a small set of principles drawn from a ‘constitution’. The supervised stage significantly improves the initial model, and gives some control over the initial behavior at the start of the RL phase, addressing potential exploration problems. The RL stage significantly improves performance and reliability.
```
jgerh, line 27: 'constitution.'

jgerh, line 27: The supervised stage significantly improves the initial model.  It gives some control over the initial behavior at the start of the RL phase, while addressing potential exploration problems.

jgerh, line 27: revisions, and

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+
+Motivation
+###############
+
```
jgerh, line 31: Constitutional AI motivation refers to designing AI systems in such a way that their objectives and behaviors are guided by a set of predefined rules or principles. It includes the following:

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+Train a CAI model
+#####################
+
+This section is a step-by-step tutorial that walks you through how to run a full CAI pipeline with a ``Mistral-7B`` LLM model. It includes the following:
+
+1. Data download and preprocessing.
+
+2. Generate responses to harmfulness prompts using a helpful-only AI assistant. Ask the model to critique its response according to a principle in the constitution, and then revise the original response in light of the critique.
+
+3. Finetune ``Mistral-7B`` with SFT on the revised responses to create a ``Mistral-7B-SL-CAI`` model.
```
jgerh, line 49: Fine-tune

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+
+4.
```
jgerh, line 51: Delete 4.

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+This section is a step-by-step tutorial that walks you through how to run a full CAI pipeline with a ``Mistral-7B`` LLM model. It includes the following:
+
+1. Data download and preprocessing.
+
+2. Generate responses to harmfulness prompts using a helpful-only AI assistant. Ask the model to critique its response according to a principle in the constitution, and then revise the original response in light of the critique.
+
+3. Finetune ``Mistral-7B`` with SFT on the revised responses to create a ``Mistral-7B-SL-CAI`` model.
+
+4. 
+   a. Use ``Mistral-7B-SL-CAI`` to generate a set of candidate responses to each prompt in a dataset of toxic prompts.
```
jgerh, line 52: replace a. with a bullet character and indent text under step 3.

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+   b. Formulate each prompt and pair into a multiple choice question, where we ask ``Mixtral-8x7B`` which response is best according to the constitution.
```
jgerh, line 53: replace b. with a bullet character and indent text under step 3.

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+   c. Blend the AI feedback preference dataset (prompts and pairs) with human feedback helpfulness dataset.
```
jgerh, line 54: replace c. with a bullet character and indent text under step 3.

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+
+5. Train a RM.
```
jgerh, line 56: Replace RM with Reward Model (RM).

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+
+6. Finetune the ``Mistral-7B-SL-CAI`` with PPO and the RM to train a ``Mistral-7B-RL-CAI`` model.
```
jgerh, line 58: Fine-tune
replace PPO with Proximal Policy Optimization (PPO)

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+
+7. Inference.
```
jgerh, line 60: 7. Run inference.

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+.. _Constitutional AI (CAI): https://arxiv.org/abs/2212.08073
+
+CAI
+###############
+The basic steps of CAI, as illustrated in Figure 1 of the paper (below):
+
+(Supervised Stage) Critique → Revision → Supervised Learning: The AI generates responses to harmfulness prompts using a helpful-only AI assistant, then critiques and revises its own responses according to a principle in the constitution, and then finetunes the original model on the revised responses.
+
+(RL Stage) AI Comparison Evaluations → Reward Model → Reinforcement Learning: The AI generates pairs of responses to harmfulness prompts using the finetuned model, then evaluates which response is better according to a principle in the constitution, and then trains a reward model based on this dataset of AI preferences and a human helpfulness preferences. The AI then trains with RL using the learned reward model.
+
```
jgerh, line 23: Add a figure title: Constitutional AI Basic Steps

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+7. Inference.
+
+.. image:: ../assets/cai_flow.png
+
+Step 1: Download models and datasets
+#############################################################################
+1. Download ``Mistral-7B-Instruct`` and ``Mistral-7B`` LLM models from https://huggingface.co/mistralai/Mistral-7B-Instruct-v0.1 and https://huggingface.co/mistralai/Mistral-7B-v0.1 into the models folder.
+
+   Then convert into .nemo format:
```
jgerh, line 68: Then,

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+      python /opt/NeMo/scripts/checkpoint_converters/convert_mistral_7b_hf_to_nemo.py --in-file /path/to/mistral-7b-instruct --out-file /models/mistral/mistral-7b-instruct.nemo
+
+
+2. Download the red teaming dataset:
+
+   .. code-block:: bash
+
+      python3 -c "from datasets import load_dataset; load_dataset('Anthropic/hh-rlhf', data_dir='red-team-attempts', split='train').to_json('/path/to/anthropic_red_team_attempts_train.json')"
+   
+   This will download the dataset to ``/path/to/anthropic_red_team_attempts_train.json``
```
jgerh, line 84: This command will download the dataset to ``/path/to/anthropic_red_team_attempts_train.json``.

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+
+
+3. Download SFT helpfulness dataset:
+   
+   .. code-block:: bash
+   
+      python3 -c "from datasets import load_dataset; load_dataset('nvidia/sft_datablend_v1', split='train').to_json('/path/to/nvidia_sft_datablend_v1_train.json')"
+
+   This will download the dataset to ``/path/to/nvidia_sft_datablend_v1_train.json``
```
jgerh, line 93: This command will download the dataset to ``/path/to/nvidia_sft_datablend_v1_train.json``.

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+      --output-filepath /path/to/cai_revisions_aligner_chat_template.jsonl \
+      --port 5999 \
+      --host <hostname or IP address of the inference service>
+
+This will generate an SL-CAI dataset of prompts and revised responses as ``cai_revisions_aligner_chat_template.json``
+
+The few-shot samples should be provided following the template in ``few_shot_samples_example.json`` (filling in the `content` tags, and choosing how many samples to use), and should include a red teaming prompt, a response from the helpful model (e.g. ``Mistral-7B`` in this tutorial), critique and revision requests and responses. The Anthropic repo outlines an example `here <https://github.com/anthropics/ConstitutionalHarmlessnessPaper/blob/main/prompts/CritiqueRevisionFewShotPrompts.json>`_.
+
+*NOTE: The tokenizer file can be found by extracting the .nemo checkpoint using `tar -xf /models/mistral/mistral-7b-Instruct.nemo`.
+There are 2 tokenizer files that end with `*.model` in the model checkpoint and they are the same, so you can use either one for data processing.*
```
jgerh, line 146: Delete * after data processing.* It shows up in the rendered text. I don't think it is needed, but pls check.

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+      --tokenizer-library sentencepiece \
+      --tokenizer-model /path/to/<TOKENIZER_FILE_NAME>.model \
+      --helpfulness-dataset-path /path/to/nvidia_sft_datablend_v1_train.json \
+      --output-filepath /path/to/cai_revisions_aligner_chat_template.jsonl \
+      --port 5999 \
+      --host <hostname or IP address of the inference service>
+
+This will generate an SL-CAI dataset of prompts and revised responses as ``cai_revisions_aligner_chat_template.json``
+
+The few-shot samples should be provided following the template in ``few_shot_samples_example.json`` (filling in the `content` tags, and choosing how many samples to use), and should include a red teaming prompt, a response from the helpful model (e.g. ``Mistral-7B`` in this tutorial), critique and revision requests and responses. The Anthropic repo outlines an example `here <https://github.com/anthropics/ConstitutionalHarmlessnessPaper/blob/main/prompts/CritiqueRevisionFewShotPrompts.json>`_.
```
jgerh, line 143: Change "The Anthropic repo outlines an example [here](https://github.com/anthropics/ConstitutionalHarmlessnessPaper/blob/main/prompts/CritiqueRevisionFewShotPrompts.json)" to "An example is shown in the Anthropic repo here:"

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+
+*NOTE: The tokenizer file can be found by extracting the .nemo checkpoint using `tar -xf /models/mistral/mistral-7b-Instruct.nemo`.
+There are 2 tokenizer files that end with `*.model` in the model checkpoint and they are the same, so you can use either one for data processing.*
+
+
+Step 3: Finetune Mistral-7B on the revised responses to create a Mistral-7B-SL-CAI model
+######################################################################################################
+
+Note that you would need to set up multi-node training run in your cluster env, depending on the type of cluster you use. For details, please refer to https://lightning.ai/docs/pytorch/stable/clouds/cluster.html
```
jgerh, line 152: Add a period at the end of the sentence: For details, please refer to https://lightning.ai/docs/pytorch/stable/clouds/cluster.html.

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+      trainer.sft.save_interval=50
+
+
+
+
+Step 4: Generate the RL-CAI (preference) dataset for RM and PPO training
+##############################################################################################################
+
+The following section runs an inference server with the SL-CAI model that we've previously trained, and queries it with red teaming prompts asking for several responses per prompt.
+These will then be ranked by a judge LLM being run from NVIDIA's NGC. An NGC API key can be acquired `here`_.
```
jgerh, line 185: The prompts will then be ranked by a judge LLM being run from NVIDIA's NGC.

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+
+The following will run the inference server:
```
jgerh, line 187: The following commands will run the inference server:

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+   python examples/nlp/cai/generate_rl_cai_dataset.py \
+      --batch-size 128 \
+      --ngc-api-key nvapi-**_YOUR-NGC-KEY-GOES-IN-HERE_** \
+      --red-teaming-file-path /path/to/anthropic_red_team_attempts_train.json \
+      --output-dir /path/to/blend_preference_dataset_with_anthropic_helpful_only \
+      --output-filename-prefix mistral_7b_cai_preference_dataset \
+      --blend-with "{'name': 'anthropic_helpful_only', 'train': {'prompts': ['/path/to/anthropic_helpful_only/anthropic_helpful_only_train_prompts_with_chat_prompt.jsonl'], 'comparisons': ['/path/to/anthropic_helpful_only/anthropic_helpful_only_train_comparisons_with_chat_prompt.jsonl']}, 'test': {'prompts': ['/path/to/anthropic_helpful_only/anthropic_helpful_only_test_prompts_with_chat_prompt.jsonl'], 'comparisons': ['/path/to/anthropic_helpful_only/anthropic_helpful_only_test_comparisons_with_chat_prompt.jsonl']}}" \
+      --port 5999
+
+This will create the ``rl-cai`` dataset files in the defined output folder with the given output filename prefix.
```
jgerh, line 221: This command will create the  . . .

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+
+
+Step 5: Train the RM
+#####################
+
```
jgerh, line 226: The following command will train the RM:

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+      exp_manager.checkpoint_callback_params.save_top_k=1 \
+      exp_manager.checkpoint_callback_params.save_nemo_on_train_end=True \
+      trainer.rm.save_interval=25 \
+      trainer.rm.val_check_interval=25 \
+      trainer.rm.limit_val_batches=100000
+
+
+The trained RM checkpoint will be saved to output dir given by ``exp_manager.explicit_log_dir``.
+
+Step 6: Finetune Mistral-7B-SL-CAI with PPO and the RM to train a Mistral-7B-RL-CAI model
```
jgerh, line 250: Fine-tune

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+      pretrained_checkpoint.restore_from_path=<path to pretrained model, e.g., mistral_7b_sl_cai_rm.nemo> \
+      trainer.ppo.inference_micro_batch_size=4 \
+      trainer.ppo.port=5567 \
+      model.offload_adam_states=True \
+      model.micro_batch_size=1 \
+      model.global_batch_size=64 \
+      model.seed=1234 \
+      exp_manager.explicit_log_dir=<path to critic output dir>
+
+Run the following command to launch actor training and reference policy server:
```
jgerh, line 266: and a reference policy server:

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+      model.optim.lr=9e-8 \
+      model.optim.sched.min_lr=9e-9 \
+      model.ppo.num_rollout_samples=512 \
+      model.ppo.rollout_micro_batch_size=8 \
+      model.ppo.length_params.max_length=1024 \
+      trainer.ppo.initial_policy_kl_penalty=0.02 \
+      remote_critic_rm.critic.ip=<ip to critic service> \
+      remote_critic_rm.critic.port=5567
+
+The trained policy (LLM) checkpoint will be saved to output dir given by ``exp_manager.explicit_log_dir``.
```
jgerh, line 288: The trained LLM policy checkpoint will be saved to the output dir given by . . .

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,308 @@
...
+           server=True \
+           tensor_model_parallel_size=8 \
+           pipeline_model_parallel_size=1 \
+           trainer.precision=bf16 \
+           trainer.devices=8 \
+           trainer.num_nodes=1 \
+           web_server=False \
+           port=1427
+
+Please wait for the server to be ready before proceeding, and follow the on-screen directive.
```
jgerh, line 308: Please wait for the server to be ready before proceeding. Then, follow the on-screen directive.

docs/user-guide/index.rst
```diff
@@ -20,6 +20,9 @@
 :ref:`Fine-tuning Stable Diffusion with DRaFT+ <model-aligner-draftp>`
    DRaFT+ is an algorithm for fine-tuning text-to-image generative diffusion models by directly backpropagating through a reward model which alleviates the mode collapse issues from DRaFT algorithm and improves diversity through regularization. 
 
+:ref:`Constitutional AI: Harmlessness from AI Feedback <model-aligner-cai>`
+   CAI is an alignment method by Anthropic allowing to incorporate AI feedback for LLM model alignment. The feedback is based on a small set of principles (Constitution) guiding the model towards the desired behavior - being helpful, honest and harmless.
```
jgerh, line 24: CAI is an alignment method by Anthropic allowing it to incorporate AI feedback for LLM model alignment.

examples/nlp/cai/sys_prompt_constitution
```diff
@@ -0,0 +1,54 @@
+You are an expert in evaluating the harmlessness of AI-generated responses. Your job is to judge responses based on their adherence to harmlessness principles.
+To make your judgment, follow the principles in the CONSTITUTION given below in triple backticks.
+
+You will be given a prompt and two or more responses to this prompt.
+Your task is to select exactly 2 responses:
+1. select the response which is most aligned with the principles in the CONSTITUTION. label this response as "chosen".
```
jgerh, line 6: 1. Select the response which is most aligned with the principles in the CONSTITUTION. Label this response as "chosen."

examples/nlp/cai/sys_prompt_constitution
```diff
@@ -0,0 +1,54 @@
...
+2. select the response which is the least aligned with the principles in the CONSTITUTION. label this response as "rejected"
```
jgerh, line 7: 2. Select the response which is the least aligned with the principles in the CONSTITUTION. Label this response as "rejected."

jgerh (commented):
I completed the copyedits to the rst files identified in [Constitutional AI: Harmlessness from AI Feedback](https://github.com/NVIDIA/NeMo-Aligner/pull/154/files#top)
#154 and submitted them for your review. Please let me know if you have any questions.

terrykong:
@gleibovich-nvidia Please review @jgerh 's copyedits

docs/user-guide/index.rst
gleibovich-nvidia: I'm not sure what does "it" is referring to here, but maybe I'm misunderstanding. 

CAI is a method which allows to incorporate AI feedback into model alignment using NeMo-Aligner. The method was invented by Anthropic. How about the following?

"CAI is an alignment method by Anthropic allowing incorporation of AI feedback for LLM model alignment

docs/user-guide/CAI.rst
gleibovich-nvidia: Thanks, revising to: 
"The basic steps of CAI are described in this section and illustrated in the figure below (Figure 1)."

docs/user-guide/CAI.rst
gleibovich-nvidia: Steps 3 and 4 are two distinct stages. I have added the missing headline for steps a. b. and c. to step 4.

docs/user-guide/CAI.rst
gleibovich-nvidia: ?
Why move the period inside the quotes?

docs/user-guide/CAI.rst
gleibovich-nvidia: The word 'here' is already a hyperlink.

docs/user-guide/CAI.rst
gleibovich-nvidia: Thanks, fixed.

docs/user-guide/CAI.rst
gleibovich-nvidia: rst doesn't support figure titles. I have tried several workaround but failed getting it to render properly. I think that the first sentence in this paragraph is sufficient. thanks.

gleibovich-nvidia:
> I completed the copyedits to the rst files identified in [Constitutional AI: Harmlessness from AI Feedback](https://github.com/NVIDIA/NeMo-Aligner/pull/154/files#top) #154 and submitted them for your review. Please let me know if you have any questions.

Thanks a lot! 
I have accepted and incorporated most of the suggestions. There are couple of comments, which I'm not sure about, to which I've replied directly.

docs/user-guide/index.rst
jgerh: Yes: "CAI is an alignment method by Anthropic allowing incorporation of AI feedback for LLM model alignment."

docs/user-guide/CAI.rst
jgerh: There are two figures in this file. If you want to refer to a figure by a number, you need to edit the figure title and add Figure 1. "<Title>" and Figure 2. "<Title>". I suggest  not adding figure numbers in this file as there are only two figures in this file and it is not necessary. 

Instead, retain my original edit: The basic steps of CAI are described in this section and illustrated in the figure below.

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,313 @@
...
+`Constitutional AI (CAI)`_ is an approach by Anthropic to train AI systems that are helpful, honest, and harmless, without relying on human feedback labels for harmlessness.
+CAI uses a set of natural language principles to guide AI behavior and self-improvement.
+The method consists of two stages: a supervised learning stage, where the AI critiques and revises its own responses according to the principles, and a reinforcement learning stage, where the AI learns from its own feedback based on the principles.
+CAI allows training a harmless, but non-evasive AI assistant that engages with harmful queries by explaining its objections to them.
+
+.. _Constitutional AI (CAI): https://arxiv.org/abs/2212.08073
+
+CAI
+###############
+The basic steps of CAI are described in this section and illustrated in the figure below (`Figure 1 <https://arxiv.org/abs/2212.08073>`_).
```
jgerh, line 18: The basic steps of CAI are described in this section and illustrated in `Figure 1 <https://arxiv.org/abs/2212.08073>`_).

docs/user-guide/CAI.rst
```diff
@@ -0,0 +1,313 @@
...
+
+(Supervised Stage) Critique → Revision → Supervised Learning: The AI generates responses to harmfulness prompts using a helpful-only AI assistant, then critiques and revises its own responses according to a principle in the constitution, and then fine-tunes the original model on the revised responses.
+
+(RL Stage) AI Comparison Evaluations → Reward Model → Reinforcement Learning: The AI generates pairs of responses to harmfulness prompts using the finetuned model, then evaluates which response is better according to a principle in the constitution, and then trains a reward model based on this dataset of AI preferences and a human helpfulness preferences. The AI then trains with RL using the learned reward model.
+
+.. image:: ../assets/cai_diagram.png
+   :alt: basic steps of the CAI process
+
+   Constitutional AI Steps. `Figure 1 <https://arxiv.org/abs/2212.08073>`_.
```
jgerh, line 27: `Figure 1. Constitutional AI Steps <https://arxiv.org/abs/2212.08073>`_.

docs/user-guide/CAI.rst
jgerh: Lines 54 to 57. Changes are good.

gleibovich-nvidia:
@terrykong FYI, we found an issue with the converted tokenizer being used within `megatron_gpt_eval.py` causing the BOS token of Mistral's prompt template not be handled correctly, when used in a conversation, rather than in a single prompt/reply. This is especially important when using few shot prompting, as the model observes an already in progress conversation (which should follow the template correctly). It seems that `megatron_gpt_eval.py` is not a good fit here as an API inference service for a conversation (chat) with several rounds of human/assistant interaction, as it does not support using HF's tokenizer natively. 

To resolve this issue, commit https://github.com/NVIDIA/NeMo-Aligner/commit/65a9c3b3531b43cee04c070185268465bcae123f is adding a lightweight inference service overriding the tokenizer to use the HF tokenizer instead, which tokenizes a conversation correctly. 
NeMo's documentation is also mentioning using the HF tokenizer [here](https://docs.nvidia.com/nemo-framework/user-guide/latest/llms/mistral/trainingpredefined.html).

docs/user-guide/CAI.rst
gleibovich-nvidia: I think I didn't explain myself well enough in the previous discussion, sorry about that. 

By referring to Figure 1, I'm not referring to Figure 1 in this specific document, but am referring to Figure 1 from CAI paper by anthropic. I.e. this is more of a citation, trying to make sure that the user understands that this is not a figure that we've made, but is taken from the paper.

See Figure 1 here. - https://arxiv.org/pdf/2212.08073.
image

How about the following to make this clearer:
The basic steps of CAI are described in this section and illustrated in the figure below (`Figure 1 <https://arxiv.org/abs/2212.08073>`_, Bai et. al).

terrykong:
> we found an issue with the converted tokenizer being used within megatron_gpt_eval.py causing the BOS token of Mistral's prompt template not be handled correctly

Is my understanding correct that the tokenizer within the megatron checkpoint does not tokenize the BOS as you expect?

Why not change the prompt template to use the BOS token that the tokenizer expects? I feel like maintaining a light fork of the megatron_gpt_eval.py should be a last resort.

docs/user-guide/CAI.rst
jgerh: OK, I understand. If we've run into the limitations of rst, then I would suggest removing the figure number altogether and just state "the figure below" or "the following figure." The same suggestion for the other figure in this file. That should work.

docs/user-guide/CAI.rst
jgerh: OK, thank you for the clarification. The text as written is correct then.

jgerh:
@gleibovich-nvidia Gal, I reviewed your comments, responded to your questions, and completed my review.

gleibovich-nvidia:
> > we found an issue with the converted tokenizer being used within megatron_gpt_eval.py causing the BOS token of Mistral's prompt template not be handled correctly
> 
> Is my understanding correct that the tokenizer within the megatron checkpoint does not tokenize the BOS as you expect?
> 
> Why not change the prompt template to use the BOS token that the tokenizer expects? I feel like maintaining a light fork of the megatron_gpt_eval.py should be a last resort.

Yes, the BOS for Mistral models is `<s>` and it is then tokenized to three different tokens `<`, `s`, `>` with the tokenizer in the megatron checkpoint, instead of tokenizing it to just one single token `<s>`. 

We prompt a **trained instruct** model (`Mistral-7B-v0.1`) asking for critiques and revisions. The model then expects this specific prompt template, considering the way it was trained and the data it has observed during SFT.

While I do agree it is not ideal to maintain code adapted from NeMo, I would like to stress that:
1.  `megatron_gpt_eval.py` is merely an example of how to run inference with `NeMo`. AFAIU, it wasn't developed to be a comprehensive solution for LLM inference, but just an example of how it can be done for specific use-cases. It doesn't implement any combination of use-cases required by potential users. This PR, on the other hand, is requiring functionality which is missing with `megatron_gpt_eval.py` to get a prompt tokenized correctly to match Mistral's expected prompt template. 
2. There are other examples within `NeMo-Aligner` where missing upstream functionality required some code adaptation and duplication. e.g. https://github.com/NVIDIA/NeMo-Aligner/blob/main/nemo_aligner/utils/train_utils.py#L16 and https://github.com/NVIDIA/NeMo-Aligner/blob/main/nemo_aligner/data/nlp/samplers.py#L28. 

There might be another way to workaround this issue that I can try on Sunday morning. Let's see how that goes. 
If all else fails, we can stick with this proposed workaround for now, and I will make sure to add a note highlighting that this code is adapted from NeMo, and we can remove this extra added functionality once this issue gets resolved in NeMo.

gleibovich-nvidia:
> @gleibovich-nvidia Gal, I reviewed your comments, responded to your questions, and completed my review.

Thanks a lot @jgerh!

jgerh (dismissed):
Completed reviews and updates from Gal Leibovich.

terrykong:
> Yes, the BOS for Mistral models is \<s\> and it is then tokenized to three different tokens \<, s, \> with the tokenizer in the megatron checkpoint, instead of tokenizing it to just one single token \<s\>.
>
> We prompt a trained instruct model (Mistral-7B-v0.1) asking for critiques and revisions. The model then expects this specific prompt template, considering the way it was trained and the data it has observed during SFT.

I feel like I'm missing something b/c shouldn't [convert_mistral_7b_hf_to_nemo.py](https://github.com/NVIDIA/NeMo/blob/main/scripts/checkpoint_converters/convert_mistral_7b_hf_to_nemo.py) add the tokenizer that the huggingface checkpoint was trained with? If for some reason you need a different one, I feel like the functionality should just be added to `convert_mistral_7b_hf_to_nemo.py`

> megatron_gpt_eval.py is merely an example of how to run inference with NeMo. AFAIU, it wasn't developed to be a comprehensive solution for LLM inference, but just an example of how it can be done for specific use-cases. It doesn't implement any combination of use-cases required by potential users. This PR, on the other hand, is requiring functionality which is missing with megatron_gpt_eval.py to get a prompt tokenized correctly to match Mistral's expected prompt template.

That's true that one script may not be comprehensive, but this example seems to just be about a tokenizer being different, so I feel like that's not a big enough reason to fork. Tokenizers are usually part of the package deal for models since swapping tokenizers can usually result in garbage when the token-ids get remapped, so I prefer to address this in the model packaging/conversion rather at inference time when swapping the tokenizer is not so much a feature, but an avenue for user error.

> There are other examples within NeMo-Aligner where missing upstream functionality required some code adaptation and duplication. e.g. https://github.com/NVIDIA/NeMo-Aligner/blob/main/nemo_aligner/utils/train_utils.py#L16 and https://github.com/NVIDIA/NeMo-Aligner/blob/main/nemo_aligner/data/nlp/samplers.py#L28.

Sure, but these are not examples to follow, but exceptions. Also, the second one even says it should be deleted.


In general, forking puts the burden on us to maintain the functionality and correctness. I would rather us focus on the algorithms here and push for Nemo/Megatron to incorporate our features when possible. If we ask and they say "no" or "not anytime soon", then I see that as a good reason for when we should fork.

docs/user-guide/CAI.rst
terrykong, line 1: Can you make this lower case? There is a new requirement that restructured text docs have to have lower case filenames

gleibovich-nvidia: Done

jgerh: IPlease check the index.rst file here https://github.com/NVIDIA/NeMo-Aligner/blob/f60ea0a6b05892da23f852c4f1dd6864cd00eeb5/docs/user-guide/index.rst and change CAI.rst to cai.rst.
cai

gleibovich-nvidia: Thanks! I've missed this one. Fixed.

docs/user-guide/index.rst
```diff
@@ -13,6 +13,7 @@
    dpo.rst
    spin.rst
    draftp.rst
+   cai.rst
 
 :ref:`Prerequisite Obtaining a Pre-Trained Model <prerequisite>`
    This section provides instructions on how to download pre-trained LLMs in .nemo format. The following section will use These base LLMs for further fine-tuning and alignment.
```
jgerh, line 19: This section provides instructions on how to download pre-trained LLMs in .nemo format. The following section will use these base LLMs for further fine-tuning and alignment.

docs/user-guide/index.rst
```diff
@@ -34,3 +35,5 @@
 
 :ref:`Fine-tuning Stable Diffusion with DRaFT+ <model-aligner-draftp>`
    DRaFT+ is an algorithm for fine-tuning text-to-image generative diffusion models by directly backpropagating through a reward model which alleviates the mode collapse issues from DRaFT algorithm and improves diversity through regularization. 
+:ref:`Constitutional AI: Harmlessness from AI Feedback <model-aligner-cai>`
+   CAI is an alignment method by Anthropic allowing incorporation of AI feedback for LLM model alignment. The feedback is based on a small set of principles (Constitution) guiding the model towards the desired behavior - being helpful, honest and harmless.
```
jgerh, line 39: Suggested revision:

CAI, an alignment method developed by Anthropic, enables the incorporation of AI feedback for aligning LLMs. This feedback is grounded in a small set of principles (referred to as the ‘Constitution’) that guide the model toward desired behaviors, emphasizing helpfulness, honesty, and harmlessness.

terrykong:
Needs https://github.com/NVIDIA/NeMo-Aligner/pull/229

docs/user-guide/index.rst
snisimov: These are parts of the document not related to CAI, i think it would be better to open a separate issue.

docs/user-guide/index.rst
snisimov: thank you we will change.

gleibovich-nvidia:
> Needs #229

Merged with `main`, and tested to work well. I think we're ready to merge.

examples/nlp/cai/generate_rl_cai_dataset.py
```diff
@@ -0,0 +1,741 @@
...
+    @staticmethod
+    def apply_prompt_with_response_template(prompt: str, response: str, system_prompt: str = ""):
+        prompt = ChatPromptTemplate.apply_prompt_template(prompt, system_prompt=system_prompt)
+        prompt_with_response = prompt + response + ChatPromptTemplate.end_signal + ChatPromptTemplate.turn_token
+        return prompt_with_response
+
+    @staticmethod
+    def extract_response(prompt: str):
+        response = prompt.rsplit(ChatPromptTemplate.assistant_message_header, 1)[-1]
+        response = response.strip().removesuffix(ChatPromptTemplate.turn_token).strip()
```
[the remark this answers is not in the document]
gleibovich-nvidia: nemo requires python >= 3.10

docs/user-guide/index.rst
terrykong: @jgerh Can you open a separate PR to address these documentation changes?

terrykong (approved):
Thanks @gleibovich-nvidia and @snisimov for working thru the review and adding CAI! LGTM!

PR #163, opened 2024-04-25
pin the seq length of dataset to the model

gshennvm:
# What does this PR do ?

make the dataset max seq length consistent with the model


odelalleau (dismissed):
lgtm but shouldn't we do the same in the spin config?

gshennvm:
> lgtm but shouldn't we do the same in the spin config?

ah forgot about that. I added it for SPIN and also added it on the SFT guide so it's less confusing for users.

odelalleau (approved):
Looks good, thanks!

PR #195, opened 2024-06-10
Implement the reward aware preference optimization algorithms.

shengyangs:
# What does this PR do ?

Implement the reward aware preference optimization algorithms.

# Changelog 

# Usage

```python
```

**Pre checks**:
- [x ] Make sure you read and followed [Contributor guidelines](/CONTRIBUTING.md)

# Additional Information
* Related to # (issue)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Next Version]
-
+- Implement reward-aware preference optimization.
+  
 ### New features and optimizations
 
 ### Breaking changes
--- a/examples/nlp/gpt/conf/gpt_dpo.yaml
+++ b/examples/nlp/gpt/conf/gpt_dpo.yaml
@@ -63,8 +63,6 @@ model:
     ref_policy_kl_penalty: 0.2
     preference_average_log_probs: False # whether normalizing log probs according to the sequence length in preference_loss
     sft_average_log_probs: ${.preference_average_log_probs} # whether normalizing log probs according to the sequence length in sft_loss
-    default_chosen_reward: 1. # the default reward for the chosen response in RPO
-    default_rejected_reward: 0. # the default reward for the rejected response in RPO
     gt_reward_scale: 1. # the scale of the rewards in RPO
     preference_loss: dpo # the preference loss, we support dpo, ipo, rpo_sq, rpo_bwd_kl, rpo_fwd_kl
     preference_loss_weight: 1 # the coefficient of the preference loss
@@ -122,7 +120,9 @@ model:
     eod_mask_loss: False # Mask loss for the end of document tokens
     index_mapping_dir: null # path to save index mapping .npy files, by default will save in the same location as data_prefix
     data_prefix: null
-
+    default_chosen_reward: 1. # the default reward for the chosen response in RPO
+    default_rejected_reward: 0. # the default reward for the rejected response in RPO
+    
   # define fields from the base model's config that should be ignored when merging with this config.
   overwrite_base_config:
     data:
--- a/examples/nlp/gpt/conf/gpt_dpo.yaml
+++ b/examples/nlp/gpt/conf/gpt_dpo.yaml
@@ -60,7 +60,14 @@ model:
   dpo:
     log_prob_forward_micro_batch_size: 1
     ref_policy_kl_penalty: 0.2
-    average_log_probs: False
+    preference_average_log_probs: False # whether normalizing log probs according to the sequence length in preference_loss
+    sft_average_log_probs: ${.preference_average_log_probs} # whether normalizing log probs according to the sequence length in sft_loss
+    default_chosen_reward: 1. # the default reward for the chosen response in RPO
+    default_rejected_reward: 0. # the default reward for the rejected response in RPO
+    gt_reward_scale: 1. # the scale of the rewards in RPO
+    preference_loss: dpo # the preference loss, we support dpo, ipo, rpo_sq, rpo_bwd_kl, rpo_fwd_kl
+    preference_loss_weight: 1 # the coefficient of the preference loss
+    sft_loss_weight: 0 # the coefficient of the SFT loss
   
   #encoder_seq_length: 4096
   #max_position_embeddings: ${model.encoder_seq_length}
--- a/nemo_aligner/data/nlp/datasets.py
+++ b/nemo_aligner/data/nlp/datasets.py
@@ -267,6 +267,10 @@ class DPOModelDataset(Dataset):
         self.reset_attention_mask = cfg.data.get("reset_attention_mask", False)
         self.eod_mask_loss = cfg.data.get("eod_mask_loss", False)
         self.eos_id = tokenizer.eos_id
+        self.default_chosen_reward = cfg.data.get("default_chosen_reward", 1.0)
+        self.default_rejected_reward = cfg.data.get("default_rejected_reward", 0.0)
+
+        self.nograd_length = 32
 
         # Checks
         assert np.min(documents) >= 0
@@ -308,9 +312,11 @@ class DPOModelDataset(Dataset):
         assert reject[0:prompt_len] == prompt, "the tokenizer for DPO has merged tokens between prompt and response"
 
         max_curr_seq_len = max(chosen_len, reject_len)
-        assert (
-            max_curr_seq_len <= self.seq_length
-        ), "tokenized text exceeds max seq len! truncate your data in preprocessing prior to DPO training"
+        if max_curr_seq_len > self.seq_length:
+            logging.warning(
+                f"WARNING: Tokenized text exceeds max seq length ({max_curr_seq_len} vs {self.seq_length})."
+                + f"The example will be ignored."
+            )
 
         chosen_tokens = torch.nn.functional.pad(
             torch.LongTensor(chosen), (0, max_curr_seq_len - chosen_len), mode="constant", value=self.eos_id
@@ -325,6 +331,15 @@ class DPOModelDataset(Dataset):
             torch.LongTensor(reject_labels), (0, max_curr_seq_len - len(reject_labels)), mode="constant", value=-100
         )
 
+        # ignore the example whose tokenized text exceeds max seq length.
+        if max_curr_seq_len > self.seq_length:
+            chosen_tokens = chosen_tokens[: self.nograd_length]
+            rejected_tokens = rejected_tokens[: self.nograd_length]
+            labels_chosen_tokens = torch.ones_like(chosen_tokens) * (-100)
+            labels_reject_tokens = torch.ones_like(rejected_tokens) * (-100)
+            chosen_len = self.nograd_length
+            reject_len = self.nograd_length
+
         output = {
             "chosen": chosen_tokens,
             "rejected": rejected_tokens,
@@ -332,6 +347,8 @@ class DPOModelDataset(Dataset):
             "rejected_length": reject_len,
             "chosen_labels": labels_chosen_tokens,
             "rejected_labels": labels_reject_tokens,
+            "chosen_reward": payload.get("chosen_reward", self.default_chosen_reward),
+            "rejected_reward": payload.get("rejected_reward", self.default_rejected_reward),
         }
         return output
 
--- a/nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
+++ b/nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
@@ -174,12 +174,18 @@ class MegatronGPTDPOModel(NLPAdapterModelMixin, MegatronGPTModel, SupervisedInte
                 )
 
                 preference_loss, acc_chosen = self.loss_func(
-                    per_token_logps, ref_logprobs, labels[:, 1:], gt_rewards, average_log_probs=self.preference_avg_log_probs
+                    per_token_logps,
+                    ref_logprobs,
+                    labels[:, 1:],
+                    gt_rewards,
+                    average_log_probs=self.preference_avg_log_probs,
                 )
 
                 sft_loss = torch.zeros_like(preference_loss)
                 if self.sft_loss_weight != 0:
-                    sft_loss = self.sft_loss_func(per_token_logps, labels[:, 1:], average_log_probs=self.sft_avg_log_probs)
+                    sft_loss = self.sft_loss_func(
+                        per_token_logps, labels[:, 1:], average_log_probs=self.sft_avg_log_probs
+                    )
                 loss = self.preference_loss_weight * preference_loss + self.sft_loss_weight * sft_loss
 
                 (
@@ -189,8 +195,9 @@ class MegatronGPTDPOModel(NLPAdapterModelMixin, MegatronGPTModel, SupervisedInte
                     reduced_acc,
                 ) = average_losses_across_data_parallel_group([loss, preference_loss, sft_loss, acc_chosen])
 
-                out_chosen, out_rejected = self.gather_and_split_rewards(per_token_logps, ref_logprobs, labels,
-                                                                         average_log_probs=self.preference_avg_log_probs)
+                out_chosen, out_rejected = self.gather_and_split_rewards(
+                    per_token_logps, ref_logprobs, labels, average_log_probs=self.preference_avg_log_probs
+                )
 
                 return (
                     loss,
--- a/nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
+++ b/nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
@@ -59,16 +59,27 @@ class MegatronGPTDPOModel(NLPAdapterModelMixin, MegatronGPTModel, SupervisedInte
         self.ref_policy_state_dict = None
 
         self.ref_policy_kl_penalty = self.cfg.dpo.get("ref_policy_kl_penalty", 0.0)
-        self.avg_log_probs = self.cfg.dpo.get("average_log_probs", False)
+        self.preference_avg_log_probs = self.cfg.dpo.get("preference_average_log_probs", False)
+        self.sft_avg_log_probs = self.cfg.dpo.get("sft_average_log_probs", self.preference_avg_log_probs)
+
+        self.preference_loss_weight = self.cfg.dpo.get("preference_loss_weight", 1)
+        self.sft_loss_weight = self.cfg.dpo.get("sft_loss_weight", 0)
+        assert (
+            self.preference_loss_weight != 0 or self.sft_loss_weight != 0
+        ), "sft loss weight and dpo loss weight cannot both be 0"
+
+        # variants of preference losses, by default DPO.
+        self.preference_loss = self.cfg.dpo.get("preference_loss", "dpo")
+        self.gt_reward_scale = self.cfg.dpo.get("gt_reward_scale", 1.0)
 
     @torch.no_grad()
-    def gather_and_split_rewards(self, pi_logprobs, ref_logprobs, labels):
+    def gather_and_split_rewards(self, pi_logprobs, ref_logprobs, labels, average_log_probs=False):
         pi_logprobs = pi_logprobs.detach()
 
         dp_group = parallel_state.get_data_parallel_group()
 
         batch_logs = self.get_reduced_masked_logps(
-            pi_logprobs - ref_logprobs, labels[:, 1:], average_log_probs=self.avg_log_probs
+            pi_logprobs - ref_logprobs, labels[:, 1:], average_log_probs=average_log_probs
         )
 
         output_list = [torch.zeros_like(batch_logs) for _ in range(dp_group.size())]
@@ -103,12 +114,14 @@ class MegatronGPTDPOModel(NLPAdapterModelMixin, MegatronGPTModel, SupervisedInte
                             "ref_policy_log_probs_rejected",
                             "chosen_labels",
                             "rejected_labels",
+                            "chosen_rewards",
+                            "rejected_rewards",
                         )
                     )
 
             batch = {key: val.cuda(non_blocking=True) if key in required_keys else None for key, val in batch.items()}
 
-            tokens, labels, ref_logprobs = None, None, None
+            tokens, labels, ref_logprobs, gt_rewards = None, None, None, None
             if batch["chosen"] is not None and batch["rejected"] is not None:
                 tokens = torch.cat((batch["chosen"], batch["rejected"]), dim=0)
 
@@ -119,6 +132,9 @@ class MegatronGPTDPOModel(NLPAdapterModelMixin, MegatronGPTModel, SupervisedInte
                 ref_logprobs = torch.cat(
                     (batch["ref_policy_log_probs_chosen"], batch["ref_policy_log_probs_rejected"]), dim=0
                 )
+
+            if batch["chosen_rewards"] is not None and batch["rejected_rewards"] is not None:
+                gt_rewards = torch.cat((batch["chosen_rewards"], batch["rejected_rewards"]), dim=0)
 
             # this is necessary if MBS > 1 with the new GBS padding logic, as you may get batch dim > 1 in some configs
             # these two lines ensure your position_ids and attn_mask are always B=1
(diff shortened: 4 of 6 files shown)


examples/nlp/gpt/conf/gpt_dpo.yaml
```diff
@@ -61,6 +61,12 @@ model:
     log_prob_forward_micro_batch_size: 1
     ref_policy_kl_penalty: 0.2
     average_log_probs: False
+    default_chosen_reward: 1.
+    default_rejected_reward: 0.
+    gt_reward_scale: 1.
+    preference_loss: dpo
```
gshennvm, line 67: can you add a comment here on the other algorithms we support?

examples/nlp/gpt/conf/gpt_dpo.yaml
```diff
@@ -61,6 +61,12 @@ model:
...
+    dpo_loss_weight: 1
```
gshennvm, line 68: should we change this to preference loss weight?

examples/nlp/gpt/conf/gpt_dpo.yaml
```diff
@@ -61,6 +61,12 @@ model:
     log_prob_forward_micro_batch_size: 1
     ref_policy_kl_penalty: 0.2
     average_log_probs: False
+    default_chosen_reward: 1.
+    default_rejected_reward: 0.
```
gshennvm, line 65: it's good to explain what these do in the config.

nemo_aligner/data/nlp/datasets.py
```diff
@@ -306,9 +308,11 @@ def __getitem__(self, idx):
...
 
         max_curr_seq_len = max(chosen_len, reject_len)
-        assert (
-            max_curr_seq_len <= self.seq_length
-        ), "tokenized text exceeds max seq len! truncate your data in preprocessing prior to DPO training"
+        if max_curr_seq_len > self.seq_length:
+            logging.warning(
+                f"WARNING: Tokenized text exceeds max seq length ({max_curr_seq_len} vs {self.seq_length})."
+                + f"The example will be ignored."
+            )
```
gshennvm, line 315: im ok with this change, but if we're going to do this how about we just do filtering in the `__init__` of this dataset?
this way we don't have to hack the `__getitem__`. Our datasets are usually not big enough for this to matter in the `__init__`

nemo_aligner/data/nlp/datasets.py
```diff
@@ -323,13 +327,32 @@ def __getitem__(self, idx):
...
+
         output = {
             "chosen": chosen_tokens,
             "rejected": rejected_tokens,
             "chosen_length": chosen_len,
             "rejected_length": reject_len,
             "chosen_labels": labels_chosen_tokens,
             "rejected_labels": labels_reject_tokens,
+            "chosen_reward": chosen_reward,
+            "rejected_reward": rejected_reward,
```
gshennvm, line 355:
```suggestion
            "chosen_reward": payload.get("chosen_reward", self.default_chosen_reward),
            "rejected_reward": payload.get("rejected_reward", self.default_rejected_reward),
```

and then no need for the if/else above

gshennvm (changes requested):
thanks for the change! just minor things here and there.

don't forget about updating the CHANGELOG and also update the DPO doc with your new variables https://github.com/NVIDIA/NeMo-Aligner/blob/main/docs/user-guide/DPO.rst

nemo_aligner/data/nlp/datasets.py
trias702: Actually, it's more efficient to do the hacking in `__getitem__` because that's the part which gets multiprocessed, but `__init__` does not. However, the current approach is rather inefficient because it takes an entire seq_length of tokens even though their labels are all negative.

examples/nlp/gpt/conf/gpt_dpo.yaml
shengyangs: Done

examples/nlp/gpt/conf/gpt_dpo.yaml
shengyangs: Done

examples/nlp/gpt/conf/gpt_dpo.yaml
shengyangs: Done

nemo_aligner/data/nlp/datasets.py
shengyangs: I haven't made any updates in this thread for the latest commit. In principle, it is more efficient if we do it in __init__ at the cost of a less prettier code.

nemo_aligner/data/nlp/datasets.py
shengyangs: Done

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
```diff
@@ -61,6 +61,16 @@ def __init__(self, cfg: DictConfig, trainer: Trainer):
         self.ref_policy_kl_penalty = self.cfg.dpo.get("ref_policy_kl_penalty", 0.0)
         self.avg_log_probs = self.cfg.dpo.get("average_log_probs", False)
 
+        self.dpo_loss_weight = self.cfg.dpo.get("dpo_loss_weight", 1)
```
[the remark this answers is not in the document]
shengyangs: Done

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
```diff
@@ -157,18 +172,34 @@ def loss_func(output_tensor):
...
+                    sft_loss = self.sft_loss_func(per_token_logps, labels[:, 1:], average_log_probs=self.avg_log_probs)
+                    loss = self.dpo_loss_weight * preference_loss + self.sft_loss_weight * sft_loss
+                else:
+                    sft_loss = torch.zeros_like(preference_loss)
+                    loss = self.dpo_loss_weight * preference_loss
+
                 reduced_loss = average_losses_across_data_parallel_group([loss])
+                reduced_preference_loss = average_losses_across_data_parallel_group([preference_loss])
+                reduced_sft_loss = average_losses_across_data_parallel_group([sft_loss])
                 reduced_acc = average_losses_across_data_parallel_group([acc_chosen])
```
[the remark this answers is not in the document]
shengyangs: Done

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
```diff
@@ -157,18 +172,34 @@ def loss_func(output_tensor):
...
+                preference_loss, acc_chosen = self.loss_func(
+                    per_token_logps, ref_logprobs, labels[:, 1:], gt_rewards, average_log_probs=self.avg_log_probs
                 )
 
+                if self.sft_loss_weight != 0:
+                    sft_loss = self.sft_loss_func(per_token_logps, labels[:, 1:], average_log_probs=self.avg_log_probs)
+                    loss = self.dpo_loss_weight * preference_loss + self.sft_loss_weight * sft_loss
+                else:
+                    sft_loss = torch.zeros_like(preference_loss)
+                    loss = self.dpo_loss_weight * preference_loss
```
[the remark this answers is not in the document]
shengyangs: Done

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
```diff
@@ -190,20 +221,62 @@ def get_reduced_masked_logps(self, logps, labels, average_log_probs=False):
...
+            ).mean(0)
+        elif self.preference_loss == "ipo":
+            loss = torch.mean((chosen_rewards - reject_rewards - 1.0 / (2.0 * self.ref_policy_kl_penalty)) ** 2, 0)
+        elif self.preference_loss == "reward_ipo":
+            chosen_gt_rewards, reject_gt_rewards = self.split_output_tensor(gt_rewards)
+            gt_rewards_delta = self.gt_reward_scale * (chosen_gt_rewards - reject_gt_rewards)
+
+            loss = torch.mean((self.ref_policy_kl_penalty * rewards_delta - gt_rewards_delta) ** 2, 0)
+        else:
+            raise NotImplementedError
```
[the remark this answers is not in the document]
shengyangs: Done

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
```diff
@@ -239,6 +312,14 @@ def get_loss_and_metrics(self, batch, forward_only):
             loss_tensors_list = [loss_reduced["avg"] for loss_reduced in losses_reduced_per_micro_batch]
             loss_tensor = torch.concat(loss_tensors_list)
             loss_mean = loss_tensor.mean()
+            sft_loss_tensors_list = [loss_reduced["avg_sft_loss"] for loss_reduced in losses_reduced_per_micro_batch]
```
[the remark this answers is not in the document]
shengyangs: I didn't update this part since there are some variables like `reward_all_mean, reward_all_std, acc` which cannot be simply put in such a loop without some less prettier if/else statements.

shengyangs:
@gshennvm @trias702 Can you take another look at the PR after the new commit to address your comments?

nemo_aligner/data/nlp/datasets.py
```diff
@@ -323,13 +327,24 @@ def __getitem__(self, idx):
             torch.LongTensor(reject_labels), (0, max_curr_seq_len - len(reject_labels)), mode="constant", value=-100
         )
 
+        # ignore the example whose tokenized text exceeds max seq length.
+        if max_curr_seq_len > self.seq_length:
+            chosen_tokens = chosen_tokens[: self.seq_length]
+            rejected_tokens = rejected_tokens[: self.seq_length]
```
trias702, line 333: Doing `chosen_tokens[: self.seq_length]` is not a good approach, because you're taking a very large amount of tokens to be used/calculated in the forward pass later on in `get_forward_output_and_loss_func`, but then the loss will be zeroed out anyway. Imagine what happens in a long context model where `seq_length` could be like 64k. I recommend you use a smaller, fixed number instead like doing:

```
chosen_tokens = chosen_tokens [:32]
rejected_tokens = rejected_tokens [:32]
```

And then shifting everything else to conform to length=32

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
```diff
@@ -157,18 +172,34 @@ def loss_func(output_tensor):
...
+                preference_loss, acc_chosen = self.loss_func(
+                    per_token_logps, ref_logprobs, labels[:, 1:], gt_rewards, average_log_probs=self.avg_log_probs
                 )
 
-                reduced_loss = average_losses_across_data_parallel_group([loss])
-                reduced_acc = average_losses_across_data_parallel_group([acc_chosen])
+                sft_loss = torch.zeros_like(preference_loss)
+                if self.sft_loss_weight != 0:
+                    sft_loss = self.sft_loss_func(per_token_logps, labels[:, 1:], average_log_probs=self.avg_log_probs)
+                loss = self.preference_loss_weight * preference_loss + self.sft_loss_weight * sft_loss
```
trias702, line 182: Have you done any testing on this logic, does it perhaps work better when the weights are tied somehow? Could you maybe test something like: `self.preference_loss_weight * preference_loss + (1.0 - self.preference_loss_weight) * sft_loss`

This would also serve to eliminate another parameter that needs to be tuned by hand via grid search. These kinds of unbounded hyperparameters are not great, because it's difficult to tune them and they have no meaningful boundedness or numeraire.

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
```diff
@@ -190,20 +221,62 @@ def get_reduced_masked_logps(self, logps, labels, average_log_probs=False):
...
+        rewards_delta = chosen_rewards - reject_rewards
+
+        if self.preference_loss == "dpo":
+            loss = -torch.nn.functional.logsigmoid(self.ref_policy_kl_penalty * rewards_delta).mean(0)
+        elif self.preference_loss == "reward_bwd_dpo":
+            logbeta_hat_chosen = torch.nn.functional.logsigmoid(self.ref_policy_kl_penalty * rewards_delta)
+            logbeta_hat_rejected = torch.nn.functional.logsigmoid(-self.ref_policy_kl_penalty * rewards_delta)
+
+            chosen_gt_rewards, reject_gt_rewards = self.split_output_tensor(gt_rewards)
+            gt_rewards_delta = self.gt_reward_scale * (chosen_gt_rewards - reject_gt_rewards)
```
trias702, line 238: How sensitive are the downstream results to this hyperparameter? Could we maybe derive it somehow rather than making it an unbounded parameter which has to be hand-tuned?

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
```diff
@@ -190,20 +221,62 @@ def get_reduced_masked_logps(self, logps, labels, average_log_probs=False):
...
+            raise NotImplementedError(f"preference_loss {self.preference_loss} is not implemented")
 
         with torch.no_grad():
             comp = chosen_rewards > reject_rewards
             acc_chosen = comp.float().mean()
 
         return loss, acc_chosen
 
+    def sft_loss_func(self, pi_logprobs, labels, average_log_probs=False):
+        logprobs = self.get_reduced_masked_logps(pi_logprobs, labels, average_log_probs=average_log_probs)
```
trias702, line 276: This just calculates a cross entropy loss right? I think it's safer to fix the `average_log_probs` then because it shouldn't ever be user changeable (as then it's no longer a cross entropy).

trias702 (changes requested):
Looks good overall, just a couple of questions/suggestions about some of the logic/params

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
```diff
@@ -190,20 +221,62 @@ def get_reduced_masked_logps(self, logps, labels, average_log_probs=False):
...
+                torch.exp(logbeta_hat_chosen) * (logbeta_hat_chosen - logalpha_hat_chosen)
+                + torch.exp(logbeta_hat_rejected) * (logbeta_hat_rejected - logalpha_hat_rejected)
+            ).mean(0)
+        elif self.preference_loss == "ipo":
+            loss = torch.mean((chosen_rewards - reject_rewards - 1.0 / (2.0 * self.ref_policy_kl_penalty)) ** 2, 0)
+        elif self.preference_loss == "reward_ipo":
+            chosen_gt_rewards, reject_gt_rewards = self.split_output_tensor(gt_rewards)
+            gt_rewards_delta = self.gt_reward_scale * (chosen_gt_rewards - reject_gt_rewards)
+
+            loss = torch.mean((self.ref_policy_kl_penalty * rewards_delta - gt_rewards_delta) ** 2, 0)
```
trias702, line 265: This is doing RPO but using L2 distance correct? Didn't you mention before that you only used reverse Bernoulli KL instead?

docs/user-guide/DPO.rst
```diff
@@ -13,6 +13,8 @@ We support both full-parameter DPO training and LoRA DPO training.
 For full-parameter DPO, there exists an actor and a reference model. Actor is initialized with the reference model and fully trainable, while reference model is frozen and used to calculate logprobs for KL-penalty loss (see `DPO paper <https://arxiv.org/pdf/2305.18290.pdf>`__). 
 For LoRA-based DPO, the actor is initialized by reference model + LoRA weights, where only the LoRA weights are trainable. Therefore, it allows us to switch between actor/reference models by simply enabling/disabling LoRA, and there is no need to store two sets of LLM weights.
 
+Besides the vanilla DPO algorithm, we support other variants of DPO algorithms including Identity preference optimization (IPO) and Reward-aware preference optimization (RPO). The algorithm is identified with the ``dpo.preference_loss`` config variable.
```
gshennvm, line 16: do we not need to label the RPO dataset with a RM? can we include instructions to do that?

if it's complicated it might make sense to make a seperate doc for it

nemo_aligner/data/nlp/datasets.py
gshennvm: agree that `__getitem__` is better since we can run it on many workers but code wise I think we should do this filtering in init and error if there's no valid samples. Then this prevents us from having to mask things, and random samples that raise this warning that might slip through

nemo_aligner/data/nlp/datasets.py
gshennvm: i think if we filter this out in init then we don't even need this logic

gshennvm (changes requested):

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
gshennvm: I believe @Kipok wanted this to be flexible in case they need to tune it, but maybe in the config we can do what you said?

Kipok: Yeah, since we don't have any experimental results yet, it's best to make it flexible and when we verify that a particular combination is the most effective, we can potentially set it as default or even remove one of the parameters

shtoshni: Hi, 

Can we have an option to control for average_log_probs for both the dpo and the SFT loss?
I have a PR in my forked repo which does that - https://github.com/NVIDIA/NeMo-Aligner/compare/main...shtoshni:NeMo-Aligner:avg_log_prob

Thanks,
Shubham

gshennvm: just for my understanding why would we need to set avg log prob = False in SFT? each sample will contribute differently to the loss that way

shtoshni: Yeah, agreed that loss averaging for SFT is a reasonable choice for most cases I can imagine. Right now the avg_log_probs param can change that. So maybe let's keep averaging true for SFT by default.

gshennvm: yeah sounds good. 8feb093655d753bdf63a487bbe59f84d8a1079d2 should fix it. I ran the CI last night and it's queued atm to check this

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
shtoshni: Moving the conversation here :) 
Can we fix the `average_log_probs` to True here?

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
shengyangs: In my opinion, the avg_log_probs should mostly be False in SFT. Looking at the SFT loss, it is 1/n sum_{i=1}^n log p(y_i | x_i), without penalizing each example based on its length.

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
gshennvm: so I had a chat with @shengyangs, and we think that it's better to keep `average_log_probs=average_log_probs`
as it is.

if we have average_log_probs = True, then both DPO and SFT loss will be scaled down by the length
if it's False then both DPO and SFT act as a sum.

so we should bind these variables together so that DPO scale is similar to SFT scale

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
shtoshni, line 1: I think tying the two doesn't allow for testing things like the Iterative RPO paper where the SFT loss is length normalized but the DPO calculations don't use the normalized log-likelihood - 
Screenshot 2024-06-13 at 4 27 05 PM

shengyangs: In this case, we need separate parameters. How about this @gshennvm ?

```
dpo.average_log_probs_preference
dpo.average_log_probs_sft
```

We set `average_log_probs_sft` default to be `average_log_probs_preference`.

shtoshni: Regarding the names, since we're using the term `sft_loss_weight`, how about `dpo_average_log_probs` and `sft_average_log_probs`?

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
shengyangs: It is quite stable. In my experiment, I just used it as 1. Depending on the ground-truth reward scales, you might want to tune it larger or smaller, but not much.

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
shengyangs: Make sense. Just a minor change. 
```
dpo.preference_average_log_probs
dpo.sft_average_log_probs
```

gshennvm: `sft_average_log_probs` and `preference_average_log_probs` sounds good to me.

maybe we should default SFT and preference to be set to the same.

what do you think?

shengyangs: Yeah, agree. We should default `sft_average_log_probs` to be the same as `preference_average_log_probs`. I will update a commit.

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
shengyangs: Yes. The reverse Bernoulli KL is called `reward_bwd_dpo`. 

I realized that the names here are kinda confusing. I will update the name to `rpo_sq, rpo_bwd_kl, rpo_fwd_kl`.

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
trias702: But if you always used 1.0, then is there any reason really to even have this parameter?If 90% of the time it's just 1.0 anyway

shengyangs: It is still useful. Imagine a case, if all the predicted rewards are in the range of 0-1  (or 0-0.1 more extremely), we want to scale up the rewards properly otherwise the policy network cannot learn. In fact, I do want to tune this hyperparameter in the upcoming ablation study.

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
shengyangs: Updated a [commit](https://github.com/NVIDIA/NeMo-Aligner/pull/195/commits/d577b176e2e9f20e546ca1ef044f5679f5ae9bbc) and made the rpo renaming.

docs/user-guide/DPO.rst
shengyangs: Updated a [commit](https://github.com/NVIDIA/NeMo-Aligner/pull/195/commits/d577b176e2e9f20e546ca1ef044f5679f5ae9bbc) and included more explanation of RPO. However, I didn't include the instruction to generate rewards, which is better covered by the reward model section instead of RPO.

nemo_aligner/data/nlp/datasets.py
shengyangs: Updated a [commit](https://github.com/NVIDIA/NeMo-Aligner/pull/195/commits/d577b176e2e9f20e546ca1ef044f5679f5ae9bbc) with changing `self.seq_length` to `32` for efficiency. I suggest that we leave the `init filtering` to a separate PR since it might need more careful discussions.

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
shengyangs: updated a [commit](https://github.com/NVIDIA/NeMo-Aligner/pull/195/commits/d577b176e2e9f20e546ca1ef044f5679f5ae9bbc) with this change.

nemo_aligner/data/nlp/datasets.py
shengyangs: I suggest that we leave the init filtering to a separate PR since it might need more careful discussions.

gshennvm: yeah that's fine cc @terrykong

nemo_aligner/data/nlp/datasets.py
```diff
@@ -331,12 +331,13 @@ def __getitem__(self, idx):
 
         # ignore the example whose tokenized text exceeds max seq length.
         if max_curr_seq_len > self.seq_length:
-            chosen_tokens = chosen_tokens[: self.seq_length]
-            rejected_tokens = rejected_tokens[: self.seq_length]
+            nograd_length = 32
```
trias702, line 334: Can you please move this to the constructor and call it self.nograd_length?

trias702 (changes requested):

nemo_aligner/data/nlp/datasets.py
shengyangs: Done in the latest commit.

trias702 (approved):
LGTM

gshennvm (approved):

CHANGELOG.md
```diff
@@ -4,7 +4,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/) and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Next Version]
-
+- Implement reward-aware preference optimization.
+  
 ### New features and optimizations
```
jgerh, line 9: capitalize headings.

### New Features and Optimizations

CHANGELOG.md
```diff
@@ -4,7 +4,8 @@ All notable changes to this project will be documented in this file.
...
 
 ### Breaking changes
```
jgerh, line 11: ### Breaking Changes

PR #202, opened 2024-06-12
Added support for float values for val_check_interval to SFT

trias702:
# What does this PR do ?

Adds support for float values for `val_check_interval` for SFT. It also adds support for floats/ints for `limit_train_batches` to SFT and DPO, as per the usage in [PTL](https://lightning.ai/docs/pytorch/stable/common/trainer.html#limit-train-batches)

This was requested by @Kipok 

# Changelog 

# Usage

```yaml
val_check_interval = 0.25   # means you will run validation 4 times per epoch
val_check_interval = 100    # means you will run validation every 100 steps of training
limit_train_batches = 0.5   # you will only use 50% of your training data per epoch
limit_train_batches = 100  # you will only consume 100 steps of your train dataloader per epoch
```

All possibilities can be used for SFT, DPO, and SPIN

**Pre checks**:

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo_aligner/utils/trainer_utils.py
+++ b/nemo_aligner/utils/trainer_utils.py
@@ -26,12 +26,12 @@ def compute_num_steps_per_epoch(
 ):
     if not sampler.drop_last:
         raise NotImplementedError("`drop_last=False` is not currently supported")
-    
+
     num_steps_per_epoch = sampler.total_samples // sampler.global_batch_size
-    
+
     if limit_train_batches is None:
         limit_train_batches = 1.0
-    
+
     if limit_train_batches >= 0 and isinstance(limit_train_batches, float):
         return min(num_steps_per_epoch, int(sampler.total_samples * limit_train_batches) // sampler.global_batch_size)
     elif limit_train_batches >= 0 and isinstance(limit_train_batches, int):
--- a/nemo_aligner/utils/trainer_utils.py
+++ b/nemo_aligner/utils/trainer_utils.py
@@ -28,17 +28,7 @@ def compute_num_steps_per_epoch(
         raise NotImplementedError("`drop_last=False` is not currently supported")
 
     num_steps_per_epoch = sampler.total_samples // sampler.global_batch_size
-
-    if limit_train_batches is None:
-        limit_train_batches = 1.0
-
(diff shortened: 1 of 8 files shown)


nemo_aligner/utils/trainer_utils.py
```diff
@@ -21,12 +21,23 @@
...
+
+    if limit_train_batches is None:
+        limit_train_batches = 1.0
+
+    if limit_train_batches >= 0 and isinstance(limit_train_batches, float):
+        return min(num_steps_per_epoch, int(sampler.total_samples * limit_train_batches) // sampler.global_batch_size)
+    elif limit_train_batches >= 0 and isinstance(limit_train_batches, int):
+        return min(num_steps_per_epoch, limit_train_batches)
+    else:
+        return num_steps_per_epoch
```
gshennvm, line 40: why can't we use the compute limit batches function below?

gshennvm: if you want to support the case of `-1` then we can edit the compute_limit_batches function for that

trias702: It doesn't really work to use the other function because they're just different enough on what they operate on to not be useful without a bunch of extra logic. The problem is that when we apply in the float case, we want to do `int(sampler.total_samples * limit_train_batches) // sampler.global_batch_size` meaning we apply the float on total samples **before** we divide by GBS, but in the other function, there's no way to easily squeeze that in.

And in terms of precision, they aren't the same:

```
int( int(7791 // 256) * 0.71 ) != int(7791 * 0.71) // 256
```

gshennvm: in the integer limit_train_batches case i think
```
compute_limit_batches(num_steps_per_epoch, limit_train_batches)
```
should be correct.

for the float case, i agree there is some precision loss but because we quantize it to an integer I'm not convinced it actually matters

in your above example

`int( int(7791 // 256) * 0.71 ) != int(7791 * 0.71) // 256` both become 21

gshennvm: as we discussed offline: 5f3746b12f825a64758792e70844de66f903054a to use compute limit batches. once CI passes will merge

gshennvm (approved):

PR #213, opened 2024-06-18
make num workers all 0 by default

gshennvm:
make num workers all 0 by default. Num workers > 0 calls a process fork/spawn underneath the hood, which causes many surprising problems for users that are trying to do things like quantization. The error is also insanely hard to debug, so I think it's better to set the default to 0.


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Breaking changes
 
 ### Bug Fixes
+- Make num workers for dataloaders 0 by default. This prevents issues when using MPI(with TRT-LLM) or more sophisticated launchers.
```
odelalleau, line 13: Just missing a space, and a minor formatting fix:

```suggestion
- Make `num_workers` for dataloaders 0 by default. This prevents issues when using MPI (with TRT-LLM) or more sophisticated launchers.
```

odelalleau (changes requested):
Just a minor tweak then good to go.

odelalleau (approved):

PR #214, opened 2024-06-18
add fix for missing extra state in rm

gshennvm:
fix issue when reward model is trained with 24.01 but loaded with 24.05

### DIFF
--- a/nemo_aligner/models/nlp/gpt/gpt_reward_model.py
+++ b/nemo_aligner/models/nlp/gpt/gpt_reward_model.py
@@ -235,5 +235,8 @@ class GPTRewardModel(GPTModel):
 
         if not self.return_rm_head_in_state_dict:
             sharded_state_dict = {k: v for k, v in sharded_state_dict.items() if "rm_head" not in k}
+        else:
+            assert self.config.fp8 is None, "fp8 is not supported for the reward model"
+            sharded_state_dict = {k: v for k, v in sharded_state_dict.items() if "rm_head._extra_state" not in k}
 
         return sharded_state_dict
--- a/nemo_aligner/models/nlp/gpt/gpt_reward_model.py
+++ b/nemo_aligner/models/nlp/gpt/gpt_reward_model.py
@@ -236,6 +236,8 @@ class GPTRewardModel(GPTModel):
         if not self.return_rm_head_in_state_dict:
             sharded_state_dict = {k: v for k, v in sharded_state_dict.items() if "rm_head" not in k}
         else:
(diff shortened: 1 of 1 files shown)


nemo_aligner/models/nlp/gpt/gpt_reward_model.py
```diff
@@ -235,5 +235,8 @@ def sharded_state_dict(self, prefix=""):
 
         if not self.return_rm_head_in_state_dict:
             sharded_state_dict = {k: v for k, v in sharded_state_dict.items() if "rm_head" not in k}
+        else:
+            assert self.config.fp8 is None, "fp8 is not supported for the reward model"
```
odelalleau, line 239: Are we guaranteed that `self.config.fp8` exists?

nemo_aligner/models/nlp/gpt/gpt_reward_model.py
```diff
@@ -235,5 +235,8 @@ def sharded_state_dict(self, prefix=""):
...
+            sharded_state_dict = {k: v for k, v in sharded_state_dict.items() if "rm_head._extra_state" not in k}
```
odelalleau, line 240: Can you add a comment to explain why we need to do this?

odelalleau (commented):
Couple of questions

nemo_aligner/models/nlp/gpt/gpt_reward_model.py
gshennvm: i looked at all offical mcore releases 0.7.0 -> 0.4.0, they all exist as a string, so this guard is ok

there is a very old commit in mcore that has this as a boolean rather than a string https://github.com/NVIDIA/Megatron-LM/blame/e5e7fe4d6c38db2a4bacb2dfb37b44a3e593a3fc/megatron/core/transformer/transformer_config.py I'm not sure what version it really corresponds to, but if you think it's a good idea we can guard against the False case as well.

nemo_aligner/models/nlp/gpt/gpt_reward_model.py
gshennvm: e17f69a5f11e60a58f10994e2ac173bbd4e39d64

nemo_aligner/models/nlp/gpt/gpt_reward_model.py
odelalleau: No that's fine, I was just worried it might break loading some previous models. Looks good to me.

odelalleau (approved):
lgtm, thanks!

Issue #227, opened 2024-07-03
Policy Log Probs and Reference Log Probs differ at 1st iteration of DPO/RPO


shengyangs:
**Describe the bug**

In DPO and all its variants, the policy is initialized at the reference policy. Therefore, in the first iteration, the log probs from the policy and the log probs from the reference policy should be exactly the same.

However, I found that the log probs differ at the 1st iteration, as shown in the figure below.

TP4 DP1. They differ.
![Screen Shot 2024-07-02 at 4 49 53 PM](https://github.com/NVIDIA/NeMo-Aligner/assets/150360624/28e8adf1-5546-41ba-bb46-25d2f539d7ae)

TP2 DP1. They are exactly the same. 
![Screen Shot 2024-07-02 at 4 47 42 PM](https://github.com/NVIDIA/NeMo-Aligner/assets/150360624/59c27a1c-1288-45a8-838a-1075d5867b16)

**Steps/Code to reproduce bug**

1. Pick a model
2. Set TP=4
3. Print out the `pi_logprobs` and `ref_logprobs` at `iteration=0`

**Expected behavior**

No matter the GBS, MBS, TP, PP, DP, Forward-MBS, they should be exactly the same.

PR #228, opened 2024-07-03
fix log probs mismatch

shengyangs:
# What does this PR do ?

This PR fixes the log probs mismatch issue (https://github.com/NVIDIA/NeMo-Aligner/issues/227) between policy and reference policy in DPO/RPO/variants.

This PR solves this issue with two changes:
* Refactor the code to compute `reference policy log probs` so that it is exactly the same as the code to compute `policy log probs`. This removes any issue related to MBS, GBS, DP, TP mismatches.
* ~~Fix [the bug here](https://github.com/NVIDIA/NeMo-Aligner/blob/8d8ef38bc7b190d9a3d0b5face76147b9e0ae863/nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py#L345) which messes up the `reference policy log probs` with their corresponding tokens.~~

With this change, the reference policy log probs match the policy log probs exactly at the first iteration. 
Screenshot 2024-07-03 at 5 39 09 PM

### DIFF
--- a/examples/nlp/gpt/conf/gpt_dpo.yaml
+++ b/examples/nlp/gpt/conf/gpt_dpo.yaml
@@ -59,6 +59,8 @@ model:
   megatron_amp_O2: True
 
   dpo:
+    # This default value ensures there are no numeric differences beween trained and reference policies when computing log probs.
+    # A higher value can be used to speed-up log probs computations, but may cause numeric differences.
     log_prob_forward_micro_batch_size: ${multiply:${model.micro_batch_size}, 2}
     ref_policy_kl_penalty: 0.2
     preference_average_log_probs: False # whether normalizing log probs according to the sequence length in preference_loss
--- a/examples/nlp/gpt/conf/gpt_dpo.yaml
+++ b/examples/nlp/gpt/conf/gpt_dpo.yaml
@@ -59,6 +59,7 @@ model:
   megatron_amp_O2: True
 
   dpo:
+    log_prob_forward_micro_batch_size: ${multiply:${model.micro_batch_size}, 2}
     ref_policy_kl_penalty: 0.2
     preference_average_log_probs: False # whether normalizing log probs according to the sequence length in preference_loss
     sft_average_log_probs: ${.preference_average_log_probs} # whether normalizing log probs according to the sequence length in sft_loss
--- a/examples/nlp/gpt/conf/gpt_dpo.yaml
+++ b/examples/nlp/gpt/conf/gpt_dpo.yaml
@@ -59,7 +59,6 @@ model:
   megatron_amp_O2: True
 
   dpo:
-    log_prob_forward_micro_batch_size: 1
     ref_policy_kl_penalty: 0.2
     preference_average_log_probs: False # whether normalizing log probs according to the sequence length in preference_loss
     sft_average_log_probs: ${.preference_average_log_probs} # whether normalizing log probs according to the sequence length in sft_loss
--- a/nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
+++ b/nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
@@ -92,7 +92,7 @@ class MegatronGPTDPOModel(NLPAdapterModelMixin, MegatronGPTModel, SupervisedInte
 
         return out_chosen.flatten(), out_rejected.flatten()
 
-    def get_forward_output_and_loss_func(self, validation_step=False):
+    def get_forward_output_and_loss_func(self, validation_step=False, logprobs_only=False):
         def fwd_output_and_loss_func(dataloader_iter, model, checkpoint_activations_all_layers=None):
             batch = next(dataloader_iter)
 
@@ -128,7 +128,10 @@ class MegatronGPTDPOModel(NLPAdapterModelMixin, MegatronGPTModel, SupervisedInte
             if batch["chosen_labels"] is not None and batch["rejected_labels"] is not None:
                 labels = torch.cat((batch["chosen_labels"], batch["rejected_labels"]), dim=0)
 
-            if batch["ref_policy_log_probs_chosen"] is not None and batch["ref_policy_log_probs_rejected"] is not None:
+            if ("ref_policy_log_probs_chosen" in batch 
+                and "ref_policy_log_probs_rejected" in batch
+                and batch["ref_policy_log_probs_chosen"] is not None 
+                and batch["ref_policy_log_probs_rejected"] is not None):
                 ref_logprobs = torch.cat(
                     (batch["ref_policy_log_probs_chosen"], batch["ref_policy_log_probs_rejected"]), dim=0
                 )
@@ -164,6 +167,13 @@ class MegatronGPTDPOModel(NLPAdapterModelMixin, MegatronGPTModel, SupervisedInte
             # so we need to explicitly cast it
             if not parallel_state.is_pipeline_last_stage():
                 output_tensor = output_tensor.to(dtype=self.autocast_dtype)
+
+            def logprobs_func(output_tensor, non_loss_data=True):
+                logprobs = from_parallel_logits_to_logprobs(
+                    vocab_parallel_logits=output_tensor, target=labels, inference_only=True, higher_stability=True,
+                )
+                
+                return {"logprobs": logprobs}
 
             def loss_func(output_tensor):
                 if validation_step and not self.cfg.data.get("validation_drop_last", True):
@@ -211,7 +221,10 @@ class MegatronGPTDPOModel(NLPAdapterModelMixin, MegatronGPTModel, SupervisedInte
                     },
                 )
 
-            return output_tensor, loss_func
+            if logprobs_only:
+                return output_tensor, logprobs_func
+            else:
+                return output_tensor, loss_func
 
         return fwd_output_and_loss_func
 
@@ -295,7 +308,7 @@ class MegatronGPTDPOModel(NLPAdapterModelMixin, MegatronGPTModel, SupervisedInte
         fwd_bwd_function = get_forward_backward_func()
 
         losses_reduced_per_micro_batch = fwd_bwd_function(
-            forward_step_func=self.get_forward_output_and_loss_func(forward_only),
+            forward_step_func=self.get_forward_output_and_loss_func(forward_only, logprobs_only=False),
             data_iterator=data_iter,
             model=self.model,
             num_microbatches=get_num_microbatches(),
@@ -384,57 +397,37 @@ class MegatronGPTDPOModel(NLPAdapterModelMixin, MegatronGPTModel, SupervisedInte
     def finish_validation_step(self):
         finish_validation_step(self)
 
-    def get_logprob_output_only_func(self, inference_only=True):
-        fwd_output_only_func = self.get_forward_output_only_func()
-
-        def log_prob_output_only_func(dataloader_iter, model):
-            batch = next(dataloader_iter)
-            logits, _ = fwd_output_only_func(iter([batch[0:3],]), model)
-
-            def id_func(logits, non_loss_data=True):
-                logprobs = from_parallel_logits_to_logprobs(
-                    vocab_parallel_logits=logits,
-                    target=batch[-1].cuda() if len(batch) == 4 else batch[0].cuda(),
-                    inference_only=inference_only,
-                    higher_stability=True,
-                )
-                return {"logprobs": logprobs}
-
-            return logits, id_func
-
-        return log_prob_output_only_func
-
     @torch.no_grad()
-    def get_logprob_batch(self, global_batch):
+    def get_logprob_batch(self, batch):
+        seq_length = batch["chosen"].shape[1]
+
+        data_iter = get_iterator_k_split(batch, get_num_microbatches())
         set_sync_funcs(self, forward_only=True)
 
-        # assumes we pad to seq length before going into the model
-        # response_tokens = sequences.cuda()
-        # labels = labels.cuda() if labels is not None else None
-
-        dp_size = parallel_state.get_data_parallel_world_size()
-        local_batch_size, seq_len = global_batch[0].shape
-        global_batch_size = local_batch_size * dp_size
-
-        forward_mbs = self.cfg.dpo.log_prob_forward_micro_batch_size
-        forward_mbs_times_dp = forward_mbs * dp_size
-
-        data_iter = get_iterator_k_split(global_batch, global_batch_size // forward_mbs_times_dp)
-
         fwd_bwd_function = get_forward_backward_func()
+
         logprobs_list = fwd_bwd_function(
-            forward_step_func=self.get_logprob_output_only_func(inference_only=True),
+            forward_step_func=self.get_forward_output_and_loss_func(logprobs_only=True),
             data_iterator=data_iter,
             model=self.model,
-            num_microbatches=global_batch_size // forward_mbs_times_dp,
+            num_microbatches=get_num_microbatches(),
             forward_only=True,
-            seq_length=seq_len,
-            micro_batch_size=forward_mbs,
+            seq_length=seq_length,
+            micro_batch_size=self.cfg.micro_batch_size
+            * 2,  # each minibatch has 2 comparisons so tensor shape will be mbs * 2
             collect_non_loss_data=True,
         )
 
         if len(logprobs_list) > 0:
-            logprobs = torch.cat([item["logprobs"] for item in logprobs_list])
+            chosen_logprobs_list = []
+            rejected_logprobs_list = []
+            for item in logprobs_list:
+                chosen_logprobs, rejected_logprobs = self.split_output_tensor(item["logprobs"])
+                chosen_logprobs_list.append(chosen_logprobs)
+                rejected_logprobs_list.append(rejected_logprobs)
+
+            logprobs = torch.cat([torch.cat(chosen_logprobs_list), 
+                                  torch.cat(rejected_logprobs_list)], dim=0)
         else:
             logprobs = None
 
@@ -449,20 +442,15 @@ class MegatronGPTDPOModel(NLPAdapterModelMixin, MegatronGPTModel, SupervisedInte
         return logprobs
 
     def get_ref_policy_logprobs(self, batch):
-        tokens = torch.cat((batch["chosen"], batch["rejected"]), dim=0)
-        masks = torch.cat((batch["attention_mask"], batch["attention_mask"]), dim=0)
-        pos_ids = torch.cat((batch["position_ids"], batch["position_ids"]), dim=0)
-        labels = torch.cat((batch["chosen_labels"], batch["rejected_labels"]), dim=0)
-        global_batch = [tokens, masks, pos_ids, labels]
-
+        
         if self.use_peft and self.ref_policy_state_dict is None:
             # when using adapters instead of full-tuning, the actor is reference model + adapters
             with adapter_control(self):
                 # With adapters disabled (meaning using the reference model), calculate ref_log_probs
-                ref_log_probs = self.get_logprob_batch(global_batch)
+                ref_log_probs = self.get_logprob_batch(batch)
         else:
             with cpu_weight_swap(self, self.ref_policy_state_dict, megatron_amp_O2=self.megatron_amp_O2):
-                ref_log_probs = self.get_logprob_batch(global_batch)
+                ref_log_probs = self.get_logprob_batch(batch)
 
         # return in GPU, trainer needs to move to cpu
         return ref_log_probs
--- a/nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
+++ b/nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
@@ -128,10 +128,8 @@ class MegatronGPTDPOModel(NLPAdapterModelMixin, MegatronGPTModel, SupervisedInte
             if batch["chosen_labels"] is not None and batch["rejected_labels"] is not None:
                 labels = torch.cat((batch["chosen_labels"], batch["rejected_labels"]), dim=0)
 
-            if ("ref_policy_log_probs_chosen" in batch 
-                and "ref_policy_log_probs_rejected" in batch
-                and batch["ref_policy_log_probs_chosen"] is not None 
-                and batch["ref_policy_log_probs_rejected"] is not None):
+            if (batch.get("ref_policy_log_probs_chosen") is not None 
+                and batch.get("ref_policy_log_probs_rejected") is not None):
                 ref_logprobs = torch.cat(
                     (batch["ref_policy_log_probs_chosen"], batch["ref_policy_log_probs_rejected"]), dim=0
                 )
@@ -180,7 +178,7 @@ class MegatronGPTDPOModel(NLPAdapterModelMixin, MegatronGPTModel, SupervisedInte
                     raise NotImplementedError("DPO does not support validation when cfg.data.drop_last=False")
 
                 per_token_logps = from_parallel_logits_to_logprobs(
-                    vocab_parallel_logits=output_tensor, target=labels, higher_stability=True
+                    vocab_parallel_logits=output_tensor, target=labels, inference_only=validation_step, higher_stability=True
                 )
 
                 preference_loss, acc_chosen = self.loss_func(
@@ -249,7 +247,7 @@ class MegatronGPTDPOModel(NLPAdapterModelMixin, MegatronGPTModel, SupervisedInte
         )
         chosen_rewards, reject_rewards = self.split_output_tensor(rewards)
         rewards_delta = chosen_rewards - reject_rewards
-
+        
         if self.preference_loss == "dpo":
             loss = -torch.nn.functional.logsigmoid(self.ref_policy_kl_penalty * rewards_delta).mean(0)
         elif self.preference_loss == "rpo_bwd_kl":
@@ -413,8 +411,7 @@ class MegatronGPTDPOModel(NLPAdapterModelMixin, MegatronGPTModel, SupervisedInte
             num_microbatches=get_num_microbatches(),
             forward_only=True,
             seq_length=seq_length,
-            micro_batch_size=self.cfg.micro_batch_size
-            * 2,  # each minibatch has 2 comparisons so tensor shape will be mbs * 2
+            micro_batch_size=self.cfg.dpo.log_prob_forward_micro_batch_size,
             collect_non_loss_data=True,
         )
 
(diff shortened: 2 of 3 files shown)


nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
```diff
@@ -165,6 +170,13 @@ def fwd_output_and_loss_func(dataloader_iter, model, checkpoint_activations_all_
             if not parallel_state.is_pipeline_last_stage():
                 output_tensor = output_tensor.to(dtype=self.autocast_dtype)
 
+            def logprobs_func(output_tensor, non_loss_data=True):
```
trias702, line 173: Why is there a `non_loss_data=True` here which is not used, but the standard `loss_func` doesn't have this param?

trias702:
The problem with the fixes in this PR is that it negates the entire purpose of the previous design: using a higher MBS for the reference log prob calculations in order to speed it up, because the reference log probs are calculated via no_grad. As I understand it, this new PR just makes the reference logprobs run with the same GBS/MBS as the actor policy, thus not taking advantage of the potential speedup from using a higher MBS for reference logprobs. Under the old code, if you set MBS == log_prob_forward_micro_batch_size did you still have the differences?

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
trias702: Shouldn't we then add this to `loss_func` as well?

shengyangs:
> The problem with the fixes in this PR is that it negates the entire purpose of the previous design: using a higher MBS for the reference log prob calculations in order to speed it up, because the reference log probs are calculated via no_grad. As I understand it, this new PR just makes the reference logprobs run with the same GBS/MBS as the actor policy, thus not taking advantage of the potential speedup from using a higher MBS for reference logprobs. Under the old code, if you set MBS == log_prob_forward_micro_batch_size did you still have the differences?

@gshennvm and I tried multiple settings in the old code and the issue persists when TP = 4

While I understand that setting a larger log_prob_forward_micro_batch_size enables potential speed up. In practice, we have been setting it to 1 or 2, which does not enjoy the benefit.

The current problem is that, we don't know if the log probs difference is due to pure randomness (I personally doubt so) or some major bug. Therefore, it is just safe to use this new change so that we can eliminate any potential issue.

Besides the code refactoring, a critical change is this https://github.com/NVIDIA/NeMo-Aligner/pull/228/files#diff-42df318806389e8adb3b04283f2bd1c33ef2a241f212cdedcc4b0a82c959e49dR423, which I believe is indeed a bug.

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
shengyangs: If `collect_non_loss_data = False`, which is true in `loss_func`, this argument `non_loss_data ` is not called.

https://github.com/NVIDIA/Megatron-LM/blob/0bc3547702464501feefeb5523b7a17e591b21fa/megatron/core/pipeline_parallel/schedules.py#L214

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
```diff
@@ -128,7 +128,12 @@ def fwd_output_and_loss_func(dataloader_iter, model, checkpoint_activations_all_
             if batch["chosen_labels"] is not None and batch["rejected_labels"] is not None:
                 labels = torch.cat((batch["chosen_labels"], batch["rejected_labels"]), dim=0)
 
-            if batch["ref_policy_log_probs_chosen"] is not None and batch["ref_policy_log_probs_rejected"] is not None:
+            if (
+                "ref_policy_log_probs_chosen" in batch
+                and "ref_policy_log_probs_rejected" in batch
+                and batch["ref_policy_log_probs_chosen"] is not None
+                and batch["ref_policy_log_probs_rejected"] is not None
```
odelalleau, line 135: Can be written more succinctly:
```suggestion
                batch.get("ref_policy_log_probs_chosen") is not None
                and batch.get("ref_policy_log_probs_rejected") is not None
```

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
odelalleau: I'm also a bit confused (like @trias702): @gshennvm is saying that we use `collect_non_loss_data=True` but @shengyangs is saying that `collect_non_loss_data = False`...

odelalleau (commented):
Only had a quick look but a few comments.

1. Instead of removing `log_prob_forward_micro_batch_size`, can we set it to default `${multiply: 2, ${..micro_batch_size}}` with a comment that says using a different value may cause a logprobs mismatch vs. the reference policy? (if that's the problem)

2. In l.184 we have
```python
                per_token_logps = from_parallel_logits_to_logprobs(
                    vocab_parallel_logits=output_tensor, target=labels, higher_stability=True
                )
```
but shouldn't it use `inference_only=validation_step`? (this is not new in this PR, but I just noticed it)

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
```diff
@@ -384,57 +399,36 @@ def prepare_for_validation_step(self):
...
         if len(logprobs_list) > 0:
-            logprobs = torch.cat([item["logprobs"] for item in logprobs_list])
+            chosen_logprobs_list = []
+            rejected_logprobs_list = []
+            for item in logprobs_list:
+                chosen_logprobs, rejected_logprobs = self.split_output_tensor(item["logprobs"])
+                chosen_logprobs_list.append(chosen_logprobs)
+                rejected_logprobs_list.append(rejected_logprobs)
+
+            logprobs = torch.cat([torch.cat(chosen_logprobs_list), torch.cat(rejected_logprobs_list)], dim=0)
```
odelalleau, line 431: So you said there was a bug here but can you explain in details what's going on? It does feel like one implementation must be off, but I find it a bit suspicious that in your new implementation it seems the order of the logprobs in the output does not match the order of the data in the input batch.
(I'd need to run it myself in debug mode to fully figure out what's going on, so I thought I'd ask first)

shengyangs:
I think I was wrong about this line. It is not a bug in the original implementation.
https://github.com/NVIDIA/NeMo-Aligner/blob/8d8ef38bc7b190d9a3d0b5face76147b9e0ae863/nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py#L345

In the original implementation, the global batch is first concatenated into an array like `array([log p(y_1^c), ... log p(y_k^c), log p(y_1^r), ..., log p(y_k^r)])` ([link](https://github.com/NVIDIA/NeMo-Aligner/blob/8d8ef38bc7b190d9a3d0b5face76147b9e0ae863/nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py#L360C9-L360C72)).  Then it is split into num_microbatches and the log probs are computed. Then the `logprobs_list` is concatenated. It is not a bug.  

In the refactored implementation in the PR, I didn't concat the chosen responses and rejected responses together like the original implementation. Instead, the batch is directly passed in without any modification. In this case, each item of the `logprobs_list` is something like `array([log p(y_1^c), log p(y_1^r)])`. Therefore, I need to split them first into `chosen & rejected` and then separately concat them together.

Overall, both implementations have no bugs. The refactored implementation in the PR enables that the log probs mismatch issue is not existent. In the original implementation, I am not sure if the mismatch is due to randomness or some parallelism issue in nemo.

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
shengyangs: https://github.com/NVIDIA/NeMo-Aligner/pull/228#issuecomment-2211098458

shengyangs:
> I'm also a bit confused (like @trias702): @gshennvm is saying that we use collect_non_loss_data=True but @shengyangs is saying that collect_non_loss_data = False...

Sorry should have made it clearer. I was saying that in `loss_func`, the `collect_non_loss_data = False`. Therefore, we don't need to have this argument (`no_loss_data`) in `loss_func`. It is only needed in `logprobs_func`.

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
shengyangs: > Instead of removing log_prob_forward_micro_batch_size, can we set it to default ${multiply: 2, ${..micro_batch_size}} with a comment that says using a different value may cause a logprobs mismatch vs. the reference policy? (if that's the problem)

This does not seem to work. @gshennvm  and I have been setting `log_prob_forward_micro_batch_size = 2` with `micro_batch_size=1`. The issue persists.


> In l.184 we have
> ```
> 
>                 per_token_logps = from_parallel_logits_to_logprobs(
>                     vocab_parallel_logits=output_tensor, target=labels, higher_stability=True
>                 )
> ```
> but shouldn't it use inference_only=validation_step? (this is not new in this PR, but I just noticed it)

Seems to me that you are right.

odelalleau: > This does not seem to work. @gshennvm and I have been setting log_prob_forward_micro_batch_size = 2 with micro_batch_size=1. The issue persists.

To be clear, I'm not saying this is the only thing to change. My point is that your new code is hardcoding
```python
micro_batch_size=self.cfg.micro_batch_size * 2
```
but I believe it'd be better to make this the default behavior while allowing one to change it if they want to (this also avoids a breaking change).

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
trias702: Also, `logprobs_func` doesn't seem to actually use `collect_non_loss_data` at all inside its function, despite having it in the signature?

trias702:
> > The problem with the fixes in this PR is that it negates the entire purpose of the previous design: using a higher MBS for the reference log prob calculations in order to speed it up, because the reference log probs are calculated via no_grad. As I understand it, this new PR just makes the reference logprobs run with the same GBS/MBS as the actor policy, thus not taking advantage of the potential speedup from using a higher MBS for reference logprobs. Under the old code, if you set MBS == log_prob_forward_micro_batch_size did you still have the differences?
> 
> @gshennvm and I tried multiple settings in the old code and the issue persists when TP = 4
> 
> While I understand that setting a larger log_prob_forward_micro_batch_size enables potential speed up. In practice, we have been setting it to 1 or 2, which does not enjoy the benefit.
> 
> The current problem is that, we don't know if the log probs difference is due to pure randomness (I personally doubt so) or some major bug. Therefore, it is just safe to use this new change so that we can eliminate any potential issue.
> 

I routinely set my `log_prob_forward_micro_batch_size` to 4 and 8 depending on the TP/DP topology of my run and I do notice a speedup. It seems like you have perhaps not been setting yours correctly to take advantage of the TP/DP differences?


Also, does this difference between reference and actor logprobs even matter? In your corrected code, have you done an A/B test with the old code on the same data, and does the fix actually improve MT-Bench performance?

shengyangs:
Thanks everyone for the comments. 

I just pushed a new commits which 
* enables `log_prob_forward_micro_batch_size` which is `2 * micro_batch_size` by default.  
* some minor changes
* add `inference_only=validation_step` in `from_parallel_logits_to_logprobs` as pointed out by @odelalleau 

With these changes, the new code can enjoy the speed up by setting `log_prob_forward_micro_batch_size` larger. Specifically, for both the original implementation and the new implementation, I tried different `log_prob_forward_micro_batch_size` and print out the first iterations of log probs. It shows that
* In the original implementation, the policy log probs and the reference log probs are different.
* In the new implementation, the policy log probs and the reference log probs are always the same for  `log_prob_forward_micro_batch_size=2,4,8`.

Therefore, I still recommend to merge this PR because 
* It can enjoy the same speed up as the original implementation by setting `log_prob_forward_micro_batch_size`.
* We are rest assured that the policy log probs and the reference log probs are no problem no matter what the underlying nemo parallelism becomes.
* @trias702 I will do an ablation study with MT bench in a training job comparing these two different implementations.


Original Implementation. `log_prob_forward_micro_batch_size=2`
original-fmbs-2

Original Implementation. `log_prob_forward_micro_batch_size=4`
original-fmbs-4

New Implementation. `log_prob_forward_micro_batch_size=2`
new-fmbs-2

New Implementation. `log_prob_forward_micro_batch_size=4`
new-fmbs-4

New Implementation. `log_prob_forward_micro_batch_size=8`
new-fmbs-8

shengyangs:
@trias702 I did an apple-to-apple comparison by running RPO on top of `15b_8T_ct3_daring-anteater_lr3e-6` (MT Bench 6.96).

With the same setting `sft_loss_weight = 1e-5`, 
* The old implementation. The best model I got is MT bench 6.91.
* The new implementation. The best model I got is MT bench 7.04.

When setting `sft_loss_weight=0.`, the new implementation gives MT bench 7.24. I didn't run the old implementation in this setting.

Overall, I think this verifies the effectiveness of the new implementation.

trias702:
> @trias702 I did an apple-to-apple comparison by running RPO on top of `15b_8T_ct3_daring-anteater_lr3e-6` (MT Bench 6.96).
> 
> With the same setting `sft_loss_weight = 1e-5`,
> 
>     * The old implementation. The best model I got is MT bench 6.91.
> 
>     * The new implementation. The best model I got is MT bench 7.04.
> 
> 
> When setting `sft_loss_weight=0.`, the new implementation gives MT bench 7.24. I didn't run the old implementation in this setting.
> 
> Overall, I think this verifies the effectiveness of the new implementation.

Did you train and generate with non-deterministic mode enabled for TransformerEngine? That alone can account for 0.2 - 0.3 swings in MT-Bench, so the 6.91 -> 7.04 may just be noise

shengyangs:
> > @trias702 I did an apple-to-apple comparison by running RPO on top of `15b_8T_ct3_daring-anteater_lr3e-6` (MT Bench 6.96).
> > With the same setting `sft_loss_weight = 1e-5`,
> > ```
> > * The old implementation. The best model I got is MT bench 6.91.
> > 
> > * The new implementation. The best model I got is MT bench 7.04.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > When setting `sft_loss_weight=0.`, the new implementation gives MT bench 7.24. I didn't run the old implementation in this setting.
> > Overall, I think this verifies the effectiveness of the new implementation.
> 
> Did you train and generate with non-deterministic mode enabled for TransformerEngine? That alone can account for 0.2 - 0.3 swings in MT-Bench, so the 6.91 -> 7.04 may just be noise

I didn't set the deterministic model for TE. It is possible that the improvement is from evaluation noise or training noise. This is interesting to me because it shows that the change causes no harm, with the additional benefit of matching the log probs.

trias702:
> > > @trias702 I did an apple-to-apple comparison by running RPO on top of `15b_8T_ct3_daring-anteater_lr3e-6` (MT Bench 6.96).
> > > With the same setting `sft_loss_weight = 1e-5`,
> > > ```
> > > * The old implementation. The best model I got is MT bench 6.91.
> > > 
> > > * The new implementation. The best model I got is MT bench 7.04.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > When setting `sft_loss_weight=0.`, the new implementation gives MT bench 7.24. I didn't run the old implementation in this setting.
> > > Overall, I think this verifies the effectiveness of the new implementation.
> > 
> > 
> > Did you train and generate with non-deterministic mode enabled for TransformerEngine? That alone can account for 0.2 - 0.3 swings in MT-Bench, so the 6.91 -> 7.04 may just be noise
> 
> I didn't set the deterministic model for TE. It is possible that the improvement is from evaluation noise or training noise. This is interesting to me because it shows that the change causes no harm, with the additional benefit of matching the log probs.

I disagree, I'm not sure this ablation study tells us anything, as we have seen multiple times that TE non-deterministic mode (especially compounded twice for training and MT-Bench generation) can cause huge swings in mtb scores. I once ran a monte carlo study and the swings can be as large as 0.4.

I am willing to go on a limb to believe that these PR changes are benign, but you cannot conduct proper scientific ablation studies when using non-deterministic code, because how will you ever know if any changes to measurement variable are due to the ablation or the non-deterministic noise? We simply cannot say with any certainty if the 6.91 -> 7.04 are caused by the PR or TE non-determinism.

gshennvm:
> I think I was wrong about this line. It is not a bug in the original implementation.
> 
> https://github.com/NVIDIA/NeMo-Aligner/blob/8d8ef38bc7b190d9a3d0b5face76147b9e0ae863/nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py#L345
> 
> In the original implementation, the global batch is first concatenated into an array like `array([log p(y_1^c), ... log p(y_k^c), log p(y_1^r), ..., log p(y_k^r)])` ([link](https://github.com/NVIDIA/NeMo-Aligner/blob/8d8ef38bc7b190d9a3d0b5face76147b9e0ae863/nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py#L360C9-L360C72)). Then it is split into num_microbatches and the log probs are computed. Then the `logprobs_list` is concatenated. It is not a bug.
> 
> In the refactored implementation in the PR, I didn't concat the chosen responses and rejected responses together like the original implementation. Instead, the batch is directly passed in without any modification. In this case, each item of the `logprobs_list` is something like `array([log p(y_1^c), log p(y_1^r)])`. Therefore, I need to split them first into `chosen & rejected` and then separately concat them together.
> 
> Overall, both implementations have no bugs. The refactored implementation in the PR enables that the log probs mismatch issue is not existent. In the original implementation, I am not sure if the mismatch is due to randomness or some parallelism issue in nemo.

ah I see -- if i understand right this does a reordering so the chosen and rejected of the same sample stays in the same batch for logprob computation

shengyangs:
> Did you train and generate with non-deterministic mode enabled for TransformerEngine? That alone can account for 0.2 - 0.3 swings in MT-Bench, so the 6.91 -> 7.04 may just be noise

@trias702  I generated responses with deterministic mode for both with sft_loss_weight=0. The original implementation gets 7.12. The new implementation gets 7.23.

> ah I see -- if i understand right this does a reordering so the chosen and rejected of the same sample stays in the same batch for logprob computation

@gshennvm  That is true. Someone this resolves the log probs mismatch issue. 

Overall, I think we should merge this PR, for 
* Log probs mismatch is not existent, even if `log_probs_forward_mbs` is big.
* You can set `log_probs_forward_mbs` big to have speed up.
* Performance-wise, a little improvement in MT bench.

trias702:
> > I think I was wrong about this line. It is not a bug in the original implementation.
> > https://github.com/NVIDIA/NeMo-Aligner/blob/8d8ef38bc7b190d9a3d0b5face76147b9e0ae863/nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py#L345
> > 
> > In the original implementation, the global batch is first concatenated into an array like `array([log p(y_1^c), ... log p(y_k^c), log p(y_1^r), ..., log p(y_k^r)])` ([link](https://github.com/NVIDIA/NeMo-Aligner/blob/8d8ef38bc7b190d9a3d0b5face76147b9e0ae863/nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py#L360C9-L360C72)). Then it is split into num_microbatches and the log probs are computed. Then the `logprobs_list` is concatenated. It is not a bug.
> > In the refactored implementation in the PR, I didn't concat the chosen responses and rejected responses together like the original implementation. Instead, the batch is directly passed in without any modification. In this case, each item of the `logprobs_list` is something like `array([log p(y_1^c), log p(y_1^r)])`. Therefore, I need to split them first into `chosen & rejected` and then separately concat them together.
> > Overall, both implementations have no bugs. The refactored implementation in the PR enables that the log probs mismatch issue is not existent. In the original implementation, I am not sure if the mismatch is due to randomness or some parallelism issue in nemo.
> 
> ah I see -- if i understand right this does a reordering so the chosen and rejected of the same sample stays in the same batch for logprob computation

Can we please verify this? At the moment this PR makes a lot of changes which aren't really necessary or even advantageous. I think we need to lock down what is the individual line(s) of code which fixes the problem, and then just change that. I am not a fan of folding the logprob function into fwd_bwd as has been done here, if instead we can just change the array order in the fwd_only logprobs function instead

trias702:
> > Did you train and generate with non-deterministic mode enabled for TransformerEngine? That alone can account for 0.2 - 0.3 swings in MT-Bench, so the 6.91 -> 7.04 may just be noise
> 
> @trias702 I generated responses with deterministic mode for both with sft_loss_weight=0. The original implementation gets 7.12. The new implementation gets 7.23.
> 
> > ah I see -- if i understand right this does a reordering so the chosen and rejected of the same sample stays in the same batch for logprob computation
> 
> @gshennvm That is true. Somehow this resolves the log probs mismatch issue.
> 
> Overall, I think we should merge this PR, for
> 
>     * Log probs mismatch is not existent, even if `log_probs_forward_mbs` is big.
> 
>     * You can set `log_probs_forward_mbs` big to have speed up.
> 
>     * Performance-wise, a little improvement in MT bench.

But this is still not correct, since you did the training with non-deterministic mode, so there is random noise in your system, and therefore we cannot isolate what is causing the positive change.

Please re-train with pure deterministic mode, and then generate with deterministic mode too, so we can all know for sure what is happening.

shengyangs:
> But this is still not correct, since you did the training with non-deterministic mode, so there is random noise in your system, and therefore we cannot isolate what is causing the positive change.
> 
> Please re-train with pure deterministic mode, and then generate with deterministic mode too, so we can all know for sure what is happening

Just checked that the models have been trained with deterministic mode.
```
&& export NVTE_ALLOW_NONDETERMINISTIC_ALGO=0 \
&& export NVTE_MASKED_SOFTMAX_FUSION=0 \
```

trias702:
> > But this is still not correct, since you did the training with non-deterministic mode, so there is random noise in your system, and therefore we cannot isolate what is causing the positive change.
> > Please re-train with pure deterministic mode, and then generate with deterministic mode too, so we can all know for sure what is happening
> 
> Just checked that the models have been trained with deterministic mode.
> 
> ```
> && export NVTE_ALLOW_NONDETERMINISTIC_ALGO=0 \
> && export NVTE_MASKED_SOFTMAX_FUSION=0 \
> ```

I'm confused, before you said you only did deterministic mode for generation, but now you are saying you used deterministic mode for generation and training both?

nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py
odelalleau: So I had a closer look to better understand what's going on:
* We use `collect_non_loss_data=True` to [get the ref policy log probs](https://github.com/NVIDIA/NeMo-Aligner/blob/dd6c271dfe7eb9831d3075417a4bd3f163c23a3c/nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py#L419), using `logprobs_only=True` which means we return the `logprobs_func()` function, which Megatron-LM [will call](https://github.com/NVIDIA/Megatron-LM/blob/0bc3547702464501feefeb5523b7a17e591b21fa/megatron/core/pipeline_parallel/schedules.py#L228) with `non_loss_data=True`
* We use `collect_non_loss_data=False` (default) to [compute training loss](https://github.com/NVIDIA/NeMo-Aligner/blob/dd6c271dfe7eb9831d3075417a4bd3f163c23a3c/nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py#L312), using `logprobs_only=False` which means we return the `loss_func()` function, which Megatron-LM [will call](https://github.com/NVIDIA/Megatron-LM/blob/0bc3547702464501feefeb5523b7a17e591b21fa/megatron/core/pipeline_parallel/schedules.py#L215) without providing the `non_loss_data` argument.

I think what's missing is something like this in `logprobs_func()`:
```python
def logprobs_func(output_tensor, non_loss_data=True):
    # This function is expected to be used only when `collect_non_loss_data=True` in Megatron-LM.
    # See https://github.com/NVIDIA/Megatron-LM/blob/0bc3547702464501feefeb5523b7a17e591b21fa/megatron/core/pipeline_parallel/schedules.py#L228
    assert non_loss_data
    logprobs = ...
```

odelalleau (commented):
Overall looks good to me. I think that if this gives the expected exact logprobs match (with no performance degradation) this is a win, even if we don't fully understand the reason of the mismatch in the first place (it'd be great to understand it, but I don't think it should be a blocker to get this merged). Just a couple of small requests to help with understanding.

shengyangs:
> I'm confused, before you said you only did deterministic mode for generation, but now you are saying you used deterministic mode for generation and training both?

There were some misunderstand between us. Here is the clarification:
* All the training were run with deterministic mode.
* Previous numbers were evaluated with non-deterministic mode. 
** `sft_loss_weight=1e-5`: original implementation: 6.91
** `sft_loss_weight=1e-5`: new implementation: 7.04
** `sft_loss_weight=0`: new implementation: 7.24
* New numbers were evaluated with deterministic mode. 
** `sft_loss_weight=0`: original implementation: 7.12
** `sft_loss_weight=0`: new implementation: 7.23

examples/nlp/gpt/conf/gpt_dpo.yaml
```diff
@@ -59,7 +59,7 @@ model:
   megatron_amp_O2: True
 
   dpo:
-    log_prob_forward_micro_batch_size: 1
+    log_prob_forward_micro_batch_size: ${multiply:${model.micro_batch_size}, 2}
```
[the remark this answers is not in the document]
shengyangs: Thanks for the suggestion. I have pushed a new commit adding this comment and also the comment regarding `non_loss_data`.

odelalleau (approved):
lgtm, thanks @shengyangs for digging into this!

shengyangs:
@gshennvm @trias702 If no more questions, can you please approve and we can merge the PR?

gshennvm (approved):

Cppowboy:
Which version of megatron-lm are you using? I found the same problem and want to use your code to quick fix this problem. But I found a little problem in finding the megatron/core/num_microbatches_calculator dependency.

Cppowboy:
The function get_ref_policy_log_prob fails when pipeline_parallel_size > 1

shengyangs:
> Which version of Megatron-LM are you using? I found the same problem and want to use your code to fix it quickly. However, I had a little problem finding the megatron.core.num_microbatches_calculator dependency.

I used `megatron_core-0.7.0`

PR #233, opened 2024-07-12
Adds support for trainer.ppo.export_rollouts_jsonl

terrykong:


terrykong:
closing, I've discovered this can be accomplished already with wandb offline dumping w/o any extra code change

PR #239, opened 2024-07-16
Update the hash of the conversion script to include TE fix for mcore conversion

terrykong:


terrykong:
closing, included an extra commit since i branched from rc branch

PR #242, opened 2024-07-18
Updates microbatch APIs to use megatron's instead of apex's

terrykong:
Same as #241 but to main

### DIFF
--- a/nemo_aligner/models/nlp/gpt/megatron_gpt_critic.py
+++ b/nemo_aligner/models/nlp/gpt/megatron_gpt_critic.py
@@ -15,7 +15,7 @@ from contextlib import nullcontext
 from enum import Enum
 
 import torch
-from megatron.core.num_microbatches_calculator import _reconfigure_microbatch_calculator, get_num_microbatches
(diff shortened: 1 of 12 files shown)


nemo_aligner/models/nlp/gpt/megatron_gpt_critic.py
```diff
@@ -15,7 +15,7 @@
 from enum import Enum
 
 import torch
-from apex.transformer.pipeline_parallel.utils import _reconfigure_microbatch_calculator, get_num_microbatches
+from megatron.core.num_microbatches_calculator import _reconfigure_microbatch_calculator, get_num_microbatches
```
gshennvm, line 18: mostly looks good to me. is there a way for them to not have a private function? can we make it an API they support?

because aligner needs to run inference at potentially a different gbs/mbs

terrykong: I've updated the mcore MR to use not use the protected syntax and updated this PR. Will see if they accept that revision

PR #253, opened 2024-07-24
Updates main Dockerfile

terrykong:
# What does this PR do ?

# Changelog 

# Usage

```python
```

**Pre checks**:

# Additional Information
* Related to # (issue)


Dockerfile
```diff
@@ -1,15 +1,26 @@
...
+ARG TE_TAG=7d576ed25266a17a7b651f2c12e8498f67e0baea
+ARG MLM_TAG=338af51452a53982d202e8386db6233adad1ce86
+ARG NEMO_TAG=6ff5bce31eecefa42d49b2c81fc57b1e1533fa7f
+ARG ALIGNER_COMMIT=db3a9ccb24e703ddf87fb6a760f3ecf71172d844
+
+FROM ${BASE_IMAGE} as named-version
+ARG APEX_TAG=master
+ARG TE_TAG=7d576ed25266a17a7b651f2c12e8498f67e0baea
+ARG MLM_TAG=core_r0.8.0
+ARG NEMO_TAG=r2.0.0rc1
```
gshennvm, line 17: im not sure we need a named version, all these tags work even if they are named too.

terrykong: Yea, I think this might be over-engineered when I think about it. The mechanism was meant to guard breaking changes going into those branches so that folks at least have a working version b/c we have no nightlies to refer to. But, assuming everyone follows code freeze, this isn't actually needed. I'll revert

PR #254, opened 2024-07-26
Enforce that mcore is >=0.8 for Aligner 0.4.0

terrykong:
This will fail until mcore pushes their wheel, but within the dockerfile that checks out ToT megatron, this should work

Depends on #252


terrykong:
Closing, since it's easier to just merge this into the dockerfile PR to avoid having a broken intermediate state

PR #279, opened 2024-09-04
Implement KTO

soumye:
# What does this PR do ?

Implement the [KTO](https://arxiv.org/abs/2402.01306) algorithm 

# Changelog 

**Pre checks**:

- [X] Does the trainer resume and restore model state all states?
- [X] Does the trainer support all parallelism techniques(PP, TP, DP)?
- [ X] Does the trainer have proper logging?

### DIFF
--- a/nemo_aligner/data/nlp/datasets.py
+++ b/nemo_aligner/data/nlp/datasets.py
@@ -353,6 +353,100 @@ class DPOModelDataset(Dataset):
         return output
 
 
+class KTOModelDataset(Dataset):
+    """This class works only with jsonl files. It assumes each line of the json file is a dictionary
+       with the prompt, along with the response (response only, no prompt), and the status denoting whether the response is
+       chosen or rejected. This Dataset will combine the prompt with the corresponding response, and then tokenize it. It 
+       will also create a score field that has 1 if the sample is chosen and 0 if rejected. It also returns the labels for 
+       each, which is the response tokens with -100 for the prompt part.
+       
+       WARNING: This class will tokenize the text, but it will raise an exception on model max seq len violations!
+                Meaning it will not truncate tokens to fit to model max seq len, because of special prefix/suffix
+                strings such as <extra_id_1>, it would not know where it is safe to truncate for each model. Therefore,
+                the user must do all truncation logic in their preprocessing step when generating the jsonl
+                used by this class. Put all special truncation logic there specific to your model.
+    """
+
+    def __init__(
+        self, cfg, tokenizer, name, data_prefix, documents, data, seq_length, seed, drop_last=True,
+    ):
+        super().__init__()
+        self.cfg = cfg
+        self.name = name
+        self.data = data
+        self.drop_last = drop_last
+        self.seq_length = seq_length
+        self.tokenizer = tokenizer
+
+        self.reset_position_ids = cfg.data.get("reset_position_ids", False)
+        self.reset_attention_mask = cfg.data.get("reset_attention_mask", False)
+        self.eod_mask_loss = cfg.data.get("eod_mask_loss", False)
+        self.eos_id = tokenizer.eos_id
+
+        np_rng = np.random.default_rng(seed=seed)
+        np_rng.shuffle(self.data)
+
+        self.nograd_length = 32
+
+        # Checks
+        assert np.min(documents) >= 0
+        assert np.max(documents) < len(self.data)
+
+    def __len__(self):
+        return len(self.data)
+
+    def encode(self, text, append_eod=False, add_dummy_prefix=True):
+        if self.cfg.data.get("apply_ftfy", False):
+            import ftfy
+
+            text = ftfy.fix_text(text)
+
+        text_ids = self.tokenizer.text_to_ids(text)
+
+        if len(text_ids) > 0 and append_eod:
+            text_ids.append(self.tokenizer.eos_id)
+
+        return text_ids, len(text_ids)
+
+    def __getitem__(self, idx):
+        """Returns a sample = prompt + response, their respective lengths, and labels.
+        Differently from DPO, we need to separate the prompt from the response.
+        """
+        payload = self.data[idx]
+        prompt, prompt_len = self.encode(payload["prompt"], append_eod=False)
+        sample, sample_len = self.encode(
+            payload["prompt"] + payload["response"], append_eod=self.cfg.data.get("append_eod", False)
+        )
+        labels = ([-100] * prompt_len) + sample[prompt_len:]
+        # Separate the response from the prompt
+        response = sample[prompt_len:]
+        preference = 1 if payload["preference"] == "chosen" else 0
+
+        assert sample[0:prompt_len] == prompt, "the tokenizer for KTO has merged tokens between prompt and response"
+
+        if sample_len > self.seq_length:
+            logging.warning(
+                f"WARNING: Tokenized text exceeds max seq length ({sample_len} vs {self.seq_length})."
+                + f"The example will be ignored."
+            )
+            # Truncate the sample and labels to the first nograd_length tokens
+            sample_len = self.nograd_length
+            sample = sample[: self.nograd_length]
+            prompt_len = self.nograd_length // 2
+            prompt = prompt[:prompt_len]
+            response = sample[prompt_len:]
+            labels = torch.ones_like(torch.LongTensor(sample)) * (-100)
+
+        output = {
+            "prompt_tokens": torch.LongTensor(prompt),
+            "response_tokens": torch.LongTensor(response),
+            "sample_length": sample_len,
+            "sample_labels": torch.LongTensor(labels),
+            "preference": preference,
+        }
+        return output
+
+
 class RegressionRewardModelDataset(RewardModelDataset):
     """This class assumes each line of the dataset file is a dictionary with "text" and "label" field, 
         where "text" is a string representing the input prompt, and "label" is a list of float or int values. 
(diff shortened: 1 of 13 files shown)


trias702:
On a first look, am a bit confused, it looks like you modified `nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py` to handle KTO by setting `preference_loss` in the config and adding the KTO loss [here](https://github.com/NVIDIA/NeMo-Aligner/blob/ff303cb481b75a192598a9e3483f47e65fef2239/nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py#L294). If this is the case, and it works correctly via the change to `megatron_gpt_dpo_model.py`, then why do we even need `nemo_aligner/models/nlp/gpt/megatron_gpt_kto_model.py`?

soumye:
So there are two versions of DPO. The one in 
`nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py` is just a modification of the DPO loss and uses pairwise preference data with both chosen and rejected. The full version of KTO which works with binary preference data ie (promt,thumps up or down) is implemented in `nemo_aligner/models/nlp/gpt/megatron_gpt_kto_model.py?`
I can remove the weaker version as strictly its functionality is all in the full version of KTO.

ertkonuk:
> So there are two versions of KTO. The one in `nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py` is just a modification of the DPO loss and uses pairwise preference data with both chosen and rejected. The full version of KTO which works with binary preference data ie (prompt,thumps up or down) is implemented in `nemo_aligner/models/nlp/gpt/megatron_gpt_kto_model.py` I can remove the weaker version as strictly its functionality is all in the full version of KTO.

I think we should remove the one in megatron_gpt_dpo_model.py. It complicates things and leads to a confusing design. As you said, we can still use the one in megatron_gpt_kto_model.py if our data is pairwise.

soumye:
> > So there are two versions of KTO. The one in `nemo_aligner/models/nlp/gpt/megatron_gpt_dpo_model.py` is just a modification of the DPO loss and uses pairwise preference data with both chosen and rejected. The full version of KTO which works with binary preference data ie (prompt,thumps up or down) is implemented in `nemo_aligner/models/nlp/gpt/megatron_gpt_kto_model.py` I can remove the weaker version as strictly its functionality is all in the full version of KTO.
> 
> I think we should remove the one in megatron_gpt_dpo_model.py. It complicates things and leads to a confusing design. As you said, we can still use the one in megatron_gpt_kto_model.py if our data is pairwise.

Sounds good. I can remote it then.

nemo_aligner/data/nlp/datasets.py
```diff
@@ -353,6 +353,91 @@ def __getitem__(self, idx):
...
+        labels = ([-100] * prompt_len) + sample[prompt_len:]
+        # Separate the response from the prompt
+        response = sample[prompt_len:]
+        preference = 1 if payload["preference"] == "chosen" else 0
+
+        assert sample[0:prompt_len] == prompt, "the tokenizer for KTO has merged tokens between prompt and response"
+
+        assert (
+            sample_len <= self.seq_length
+        ), "tokenized text exceeds max seq len! truncate your data in preprocessing prior to KTO training"
```
trias702, line 429: I would recommend you amend this part to do the truncation automatically similar to how we now do it for DPO: https://github.com/NVIDIA/NeMo-Aligner/blob/f1fa2dcaee9b19deee0d2d33b9b6e3251ab7bac7/nemo_aligner/data/nlp/datasets.py#L335

nemo_aligner/models/nlp/gpt/megatron_gpt_kto_model.py
```diff
@@ -0,0 +1,414 @@
...
+
+    def prepare_for_validation_step(self):
+        prepare_for_validation_step(self)
+
+    def finish_validation_step(self):
+        finish_validation_step(self)
+
+    @torch.no_grad()
+    def get_logprob_batch(self, batch):
+        seq_length = batch["samples"].shape[1]
```
trias702, line 362: Add: `batch_size = batch["chosen"].shape[0]`

nemo_aligner/models/nlp/gpt/megatron_gpt_kto_model.py
```diff
@@ -0,0 +1,414 @@
...
+
```
trias702, line 363: Add: `num_microbatches = divide(batch_size, self.cfg.kto.log_prob_forward_micro_batch_size)`

nemo_aligner/models/nlp/gpt/megatron_gpt_kto_model.py
```diff
@@ -0,0 +1,414 @@
...
+        data_iter = get_iterator_k_split(batch, get_num_microbatches())
```
trias702, line 364: Should be: `data_iter = get_iterator_k_split(batch, num_microbatches)`

nemo_aligner/models/nlp/gpt/megatron_gpt_kto_model.py
```diff
@@ -0,0 +1,414 @@
...
+        set_sync_funcs(self, forward_only=True)
+
+        fwd_bwd_function = get_forward_backward_func()
+
+        logprobs_list = fwd_bwd_function(
+            forward_step_func=self.get_forward_output_and_loss_func(logprobs_only=True),
+            data_iterator=data_iter,
+            model=self.model,
+            num_microbatches=get_num_microbatches(),
```
trias702, line 373: Change to: `num_microbatches=num_microbatches`

nemo_aligner/models/nlp/gpt/megatron_gpt_kto_model.py
```diff
@@ -0,0 +1,414 @@
...
+            forward_only=True,
+            seq_length=seq_length,
+            micro_batch_size=self.cfg.kto.log_prob_forward_micro_batch_size,
```
trias702, line 376: Change to: `micro_batch_size=self.cfg.kto.log_prob_forward_micro_batch_size * 2,`

trias702 (changes requested):
LGTM, just a few small changes

nemo_aligner/models/nlp/gpt/megatron_gpt_kto_model.py
soumye: In KTO there shoulndn't be `* 2`  in the `micro_batch_size` as there is only one response per prompt over which we compute the loss.

nemo_aligner/models/nlp/gpt/megatron_gpt_kto_model.py
soumye: In KTO as there is only one response per prompt over which we compute the loss. So `self.cfg.kto.log_prob_forward_micro_batch_size=1` anyways so it wouldn't matter. But changing it for consistency with DPO code

nemo_aligner/data/nlp/datasets.py
soumye: Added

nemo_aligner/models/nlp/gpt/megatron_gpt_kto_model.py
trias702: Ah yes, that's a fair point, in that case, no need to change it, but just a comment to remind people it's just a single response per prompt and not like DPO

nemo_aligner/models/nlp/gpt/megatron_gpt_kto_model.py
```diff
@@ -0,0 +1,418 @@
...
+    def finish_validation_step(self):
+        finish_validation_step(self)
+
+    @torch.no_grad()
+    def get_logprob_batch(self, batch):
+        seq_length = batch["samples"].shape[1]
+        batch_size = batch["samples"].shape[0]
+
+        # num_microbatches = get_num_microbatches()
+        num_microbatches = divide(batch_size, self.cfg.kto.log_prob_forward_micro_batch_size)
```
trias702, line 367: Should be: `divide(batch_size, min(batch_size, self.cfg.kto.log_prob_forward_micro_batch_size))`

nemo_aligner/models/nlp/gpt/megatron_gpt_kto_model.py
```diff
@@ -0,0 +1,418 @@
...
+        fwd_bwd_function = get_forward_backward_func()
+
+        logprobs_list = fwd_bwd_function(
+            forward_step_func=self.get_forward_output_and_loss_func(logprobs_only=True),
+            data_iterator=data_iter,
+            model=self.model,
+            num_microbatches=num_microbatches,
+            forward_only=True,
+            seq_length=seq_length,
+            micro_batch_size=self.cfg.kto.log_prob_forward_micro_batch_size,
```
trias702, line 380: Should be: `micro_batch_size=min(batch_size, self.cfg.kto.log_prob_forward_micro_batch_size),`

Just a guard in case someone enters a value too high for the forward mbs

trias702 (changes requested):
LGTM, just one last item

nemo_aligner/models/nlp/gpt/megatron_gpt_kto_model.py
soumye: Good idea.

nemo_aligner/models/nlp/gpt/megatron_gpt_kto_model.py
soumye: done

trias702 (approved):
LGTM, thanks

PR #280, opened 2024-09-05
fix micro batch size logic in DPO

gshennvm:
# What does this PR do ?

fix mbs logic in dpo


trias702:
We should test that this doesn't change models trained using the old logic vs new logic, and also that reference logprobs == actor logprobs for the first epoch

terrykong (approved):

PR #285, opened 2024-09-06
ci: Add build job

ko3n1g:
# What does this PR do ?

* GitHub Workflow:
  * Uses caching by pulling from an image produced in main and from an image based of the current PR. The latter helps for cases where the PR bumps dependencies that diverge from main
  * These caching strategies are not yet relevant since we only use a single builder node, once we scale to two nodes this will be highly benefitial given the expensive build time
  * Since we want NeMo-aligner to be "up-to-date", i feed the current sha as build arg into the build
* Dockerfile:
  * To optimize layer caching I move build args closer to their usage

There's a lot to do or at least to experiment with to reduce build time and size. Some things I have in mind:

* Moving less expensive installs like mcore, nemo, and nemo-aligner to the bottom
* Parallelize with multi-stage: i think this is the most difficult since we need to know which files to copy over. But also highest gain in terms of size and build time optimization
* Easier but also less impact: Reducing number of layers to reduce size (since we avoid duplicated compilers/build-tools that each overlayfs directory otherwise needs to track)

However, i don't think these optimizations are trivial so the next step would be focusing on test workflow before circling back to the build process. 

# Changelog 

# Usage

```python
```

**Pre checks**:

# Additional Information
* Related to # (issue)


Dockerfile
```diff
@@ -1,15 +1,7 @@
...
 FROM ${BASE_IMAGE}
-ARG APEX_TAG=59b80ee8df79cec125794949327f29913c328746
-ARG TE_TAG=7d576ed25266a17a7b651f2c12e8498f67e0baea
-ARG MLM_TAG=a3fe0c75df82218901fa2c3a7c9e389aa5f53182  # On: core_r0.8.0
-ARG NEMO_TAG=e033481e26e6ae32764d3e2b3f16afed00dc7218  # On: r2.0.0rc1
-ARG ALIGNER_COMMIT=main
 
-ARG PYTRITON_VERSION=0.5.10
-ARG PROTOBUF_VERSION=4.24.4
-ARG TRTLLM_VERSION=v0.10.0
```
ko3n1g, line 12: Actually we could revert this change. I studied the books again and cache invalidation happens only where ARGS are used, not where they are defined. So it would be fine to define all collectively at the top. I guess there are different tastes (some like to define it close to first usage, others as a "ToC" collectively). I'm fine with both, let me know what you'd prefer

terrykong: I'd prefer at the top. I think it's easier to inspect

terrykong: I double checked I could pull the image, so I think once that's reverted we can merge

PR #290, opened 2024-09-18
feat: update dockerfile to allow incremental builds and change TRTLLM error to point to dockerfile

terrykong:
# What does this PR do ?

This PR contains two changes:
1. Change to the dockerfile to allow two types of builds
```sh
# Original build
docker buildx build -t aligner:latest .

# An incremental build where only nemo-aligner gets bumped changes

docker buildx build --target=aligner-bump \
  --build-arg=BASE_IMAGE=nvcr.io/nvidia/nemo:24.07 \
  --build-arg=ALIGNER_COMMIT=dev \
  -t aligner:latest .
```

2. Moves the error message when either (a) TRTLLM isn't installed (b) doesn't have v10 refit functionality into GenerateGPTLLM and have the message point to this new dockerfile

# Changelog 

# Usage

```python
```

**Pre checks**:

# Additional Information
* Related to # (issue)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,12 @@
+# To build NeMo-Aligner from a base PyTorch container:
+#
+#   docker buildx build -t aligner:latest .
+#
+# To update NeMo-Aligner from a pre-built NeMo-Framework container:
+#
+#   docker buildx build --target=aligner-bump --build-arg=BASE_IMAGE=nvcr.io/nvidia/nemo:24.07 -t aligner:latest .
+#
+
 # Number of parallel threads for compute heavy build jobs
 # if you get errors building TE or Apex, decrease this to 4
 ARG MAX_JOBS=8
@@ -12,16 +21,33 @@ ARG PROTOBUF_VERSION=4.24.4
 
 ARG BASE_IMAGE=nvcr.io/nvidia/pytorch:24.03-py3
 
-FROM ${BASE_IMAGE}
+FROM ${BASE_IMAGE} AS aligner-bump
 
-ARG MAX_JOBS
-
-# needed in case git complains that it can't detect a valid email, this email is fake but works
-RUN git config --global user.email "worker@nvidia.com"
+ARG ALIGNER_COMMIT
 
 WORKDIR /opt
 
+# NeMo Aligner
+RUN <<"EOF" bash -exu
+if [[ ! -d NeMo-Aligner ]]; then
+    git clone https://github.com/NVIDIA/NeMo-Aligner.git
+    cd NeMo-Aligner
+    git checkout $ALIGNER_COMMIT
+    pip install --no-deps -e .
+    cd -
+fi
+cd NeMo-Aligner
+git fetch -a
+git checkout -f ${ALIGNER_COMMIT}
+git pull
+EOF
+
+FROM aligner-bump as final
+
+# needed in case git complains that it can't detect a valid email, this email is fake but works
+RUN git config --global user.email "worker@nvidia.com"
 # install TransformerEngine
+ARG MAX_JOBS
 ARG TE_TAG
 RUN pip uninstall -y transformer-engine && \
     git clone https://github.com/NVIDIA/TransformerEngine.git && \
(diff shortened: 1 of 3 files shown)


terrykong:
Push this into dev instead

Dockerfile
```diff
@@ -1,24 +1,48 @@
...
+    git clone https://github.com/NVIDIA/NeMo-Aligner.git
+    cd NeMo-Aligner
+    git checkout $ALIGNER_COMMIT
+    pip install --no-deps -e .
+    cd -
+fi
+cd NeMo-Aligner
+git fetch -a
+git checkout -f ${ALIGNER_COMMIT}
+git pull
```
ko3n1g, line 27: Any reason why not a `pip install --no-deps -e git+https://github.com/NVIDIA/NeMo-Aligner.git$ALIGNER_COMMIT` ?

Dockerfile
```diff
@@ -1,24 +1,48 @@
...
 # if you get errors building TE or Apex, decrease this to 4
 ARG MAX_JOBS=8
+# Git refs for dependencies
+ARG TE_TAG=7d576ed25266a17a7b651f2c12e8498f67e0baea
+ARG APEX_TAG=59b80ee8df79cec125794949327f29913c328746
+ARG PYTRITON_VERSION=0.5.10
+ARG NEMO_TAG=e033481e26e6ae32764d3e2b3f16afed00dc7218  # On: r2.0.0rc1
+ARG MLM_TAG=a3fe0c75df82218901fa2c3a7c9e389aa5f53182  # On: core_r0.8.0
+ARG TRTLLM_VERSION=v0.10.0
+ARG PROTOBUF_VERSION=4.24.4
```
ko3n1g, line 42: Should we keep the logic of maintaining all build args at the very top? I found your preference good tbh

ko3n1g (changes requested):
Two minor improvements that I would suggest for better readability

Dockerfile
ko3n1g: Actually, nevermind. Docker is not smart enough to understand when exactly an `ARG` is being _used_. So whenever there's a `RUN` all previously defined `ARG`  will be injected as environment variables and so can destroy the cache. Even though it like it from an overview-perspective, it's not good from a caching perspective to define all at top

terrykong: I see. Since your #301 will address this, let's keep it this way just for 0.5 to minimize churn and then land the ARG restructure in 301

Dockerfile
terrykong: Great suggestion. I can try that

terrykong: Hm, so there are some weird nuances with using VCS editable installs.

1. looks like the directory will be the package name (nemo_aligner) and not the repository name (NeMo-Aligner)
2. repeated installs updates the local branch to the new remote ref, which appears to be equivalent to `git reset --hard origin/$new_ref`. This means when you `cd` into the cloned copy, it looks off:
```
root@dbc97c04ea58:/tmp/nemo-aligner# git status
On branch dev
Your branch and 'origin/dev' have diverged,
and have 241 and 28 different commits each, respectively.
  (use "git pull" to merge the remote branch into yours)

nothing to commit, working tree clean
```

I think for now we should just land this and think about a better way to do this afterwards. For posterity, we need to do something like this:

```sh
 pip install --src=/opt --no-deps -e git+https://github.com/NVIDIA/NeMo-Aligner.git@${ALIGNER_COMMIT}#egg=nemo_aligner
```

ko3n1g (approved):

Dockerfile
```diff
@@ -12,16 +21,33 @@ ARG PROTOBUF_VERSION=4.24.4
...
+FROM ${BASE_IMAGE} AS aligner-bump
 
-# needed in case git complains that it can't detect a valid email, this email is fake but works
-RUN git config --global user.email "worker@nvidia.com"
+ARG ALIGNER_COMMIT
 
 WORKDIR /opt
 
+# NeMo Aligner
+RUN <<"EOF" bash -exu
```
gshennvm, line 31: sorry, why not keep the old

```
RUN git clone https://github.com/NVIDIA/NeMo-Aligner.git && \
    cd NeMo-Aligner && \
    git pull && \
    if [ ! -z $ALIGNER_COMMIT ]; then \
        git fetch origin $ALIGNER_COMMIT && \
        git checkout FETCH_HEAD; \
    fi && \
    pip install --no-deps -e .
```
what is wrong with that approach?

gshennvm: ah i see! this essentially prevents us from overwriting the old aligner

Dockerfile
```diff
@@ -12,16 +21,33 @@ ARG PROTOBUF_VERSION=4.24.4
 
 ARG BASE_IMAGE=nvcr.io/nvidia/pytorch:24.03-py3
 
-FROM ${BASE_IMAGE}
-
-ARG MAX_JOBS
+FROM ${BASE_IMAGE} AS aligner-bump
```
gshennvm, line 24: if the user starts from a 24.07 image, wouldn't it still reinstall everything else other than aligner?

Dockerfile
terrykong: yea, exactly

gshennvm (approved):

PR #320, opened 2024-09-26
feat: Upgrading TRTLLM to v13

terrykong:
# What does this PR do ?

This is a draft of the changes required to move Aligner to TRTLLM v12. There are changes in Nemo/TRTLLM that need to land before this one:

* NeMo:
    * https://github.com/NVIDIA/NeMo/pull/10654
    * https://github.com/NVIDIA/NeMo/pull/10651
    * https://github.com/NVIDIA/NeMo/pull/10652
    * ~~https://github.com/NVIDIA/NeMo/pull/10653~~
        * Removing from required list since this was more of a QoL fix for debugging, but not necessary for this PR
    * https://github.com/NVIDIA/NeMo/pull/10863
* TRTLLM:
    * in-flight changes in: https://github.com/NVIDIA/NeMo-Aligner/blob/terryk/trtllm-v12/setup/trtllm.patch 

# Changelog 

# Usage

```python
```

**Pre checks**:

# Additional Information
* Related to # (issue)


nemo_aligner/utils/distributed.py
```diff
@@ -82,6 +86,53 @@ def broadcast_2d_tensor(tensor, src, group, dtype=torch.float32):
...
+    All the ranks that send or receive data must call this function.
+    
+    dtype: Optional dtype, {dir(torch)}
+    """
+    if torch.distributed.get_rank() == src:
+        tensor = tensor.cuda()
+        if dtype:
+            tensor = tensor.to(dtype)
+
+        input_dtype_enum = torch.tensor(_DTYPE_TO_ENUM.get(tensor.dtype, 0), device="cuda", dtype=torch.int32)
```
gshennvm, line 112: is there a way to get dtype to int automatically using torch?

terrykong: not that i'm aware of. if you know some util let me know.

the only thing i found in pytorch src itself was [this](https://github.com/pytorch/pytorch/blob/46f158bfbc5c3d6020f8458ae749b89addc7ffa8/torch/_export/converter.py#L194), but it's protected, so I don't know if we want to depend on a non-public API

terrykong:
I tested v13 locally and actually it required no code update! I'll change this PR to jump straight to V13

.github/workflows/cicd-main.yml
```diff
@@ -28,6 +36,60 @@ concurrency:
...
+    if: contains(fromJSON(needs.pre-flight.outputs.test_to_run), 'functional') || needs.pre-flight.outputs.all == 'true'
+    strategy:
+     matrix:
+       include:
+        # ALIGNER_CI_DIR/ALIGNER_REPO_DIR is available in the shell env
+        - TEST_NAME: ppo
+          SCRIPT: |
+            RM_NEMO_FILE=${ALIGNER_CI_DIR}/checkpoints/llama3--nlayers4-hidden64-ffn224-dummy_rm-megatron_gpt.nemo \
+            ACTOR_NEMO_FILE=${ALIGNER_CI_DIR}/checkpoints/tiny-llama3-results-nlayers2-hidden128-ffn448-nhead4-qgroup2-megatron_gpt.nemo \
+            bash ${ALIGNER_REPO_DIR}/tests/functional/ppo.sh
```
ko3n1g, line 85: WDYT about storing this shell script separately in the repo? That would allow running/testing it easily on other infra too

.github/workflows/_run_test.yml
```diff
@@ -0,0 +1,102 @@
...
+            ${{ inputs.SCRIPT }}
+            RUN_TEST_EOF
+            )
+              #docker run --rm --runtime=nvidia --gpus all --shm-size=8g --env TRANSFORMERS_OFFLINE=0 --env HYDRA_FULL_ERROR=1 --volume /mnt/datadrive/TestData:/home/TestData:ro nemoci.azurecr.io/nemo_aligner_container:${{ github.run_id }} bash -eux -o pipefail -c "$cmd"
+              docker run --rm --runtime=nvidia --gpus all --shm-size=8g \
+                --env TRANSFORMERS_OFFLINE=0 --env HYDRA_FULL_ERROR=1 --env HF_HOME=/home/TestData/aligner/hf_home --env ALIGNER_CI_DIR=/home/TestData/aligner --env ALIGNER_REPO_DIR=/opt/NeMo-Aligner \
+                --volume /mnt/datadrive/TestData/aligner/checkpoints:/home/TestData/aligner/checkpoints:ro \
+                --volume /mnt/datadrive/TestData/aligner/hf_home/hub:/home/TestData/aligner/hf_home/hub:ro \
+                nemoci.azurecr.io/nemo_aligner_container:${{ github.event.pull_request.number }} \
+                bash -eux -o pipefail -c "$cmd"
```
ko3n1g, line 78: Dont need to change it in this PR, but it's unsafe to use `docker run` with GitHub actions & our infra. See other discussion

.github/workflows/_run_test.yml
```diff
@@ -0,0 +1,102 @@
...
+            ${{ inputs.AFTER_SCRIPT }}
+            RUN_TEST_EOF
+            )
+            #docker run --rm --runtime=nvidia --gpus all --shm-size=8g --env TRANSFORMERS_OFFLINE=0 --env HYDRA_FULL_ERROR=1 --volume /mnt/datadrive/TestData:/home/TestData nemoci.azurecr.io/nemo_aligner_container:${{ github.run_id }} bash -eux -o pipefail -c "$cmd"
+            docker run --rm --runtime=nvidia --gpus all --shm-size=8g \
+              --env TRANSFORMERS_OFFLINE=0 --env HYDRA_FULL_ERROR=1 --env HF_HOME=/home/TestData/aligner/hf_home --env ALIGNER_CI_DIR=/home/TestData/aligner \
+              --volume /mnt/datadrive/TestData/aligner/checkpoints:/home/TestData/aligner/checkpoints:ro \
+              --volume /mnt/datadrive/TestData/aligner/hf_home/hub:/home/TestData/aligner/hf_home/hub:ro \
+              nemoci.azurecr.io/nemo_aligner_container:${{ github.event.pull_request.number }} \
+              bash -eux -o pipefail -c "$cmd"
```
ko3n1g, line 102: https://github.com/NVIDIA/NeMo/blob/main/.github/workflows/_test_template.yml#L54

ko3n1g (commented):
as discussed

.gitignore
```diff
@@ -154,4 +154,3 @@ wandb
 
 # Checkpoints, config files and temporary files created in tutorials.
 .hydra/
-*.jsonl
```
gshennvm, line 157: just making sure -- is this because we want to add some jsonl files on the unit test?

tests/test_mcore_utilities.py
```diff
@@ -0,0 +1,99 @@
...
+    def initialize_distributed():
+        if not torch.distributed.is_initialized() and Utils.rank >= 0:
+            print(f"Initializing torch.distributed with rank: {Utils.rank}, world_size: {Utils.world_size}")
+            torch.cuda.set_device(Utils.rank % torch.cuda.device_count())
+            init_method = "tcp://"
+            master_ip = os.getenv("MASTER_ADDR", "localhost")
+            master_port = os.getenv("MASTER_PORT", "6000")
+            init_method += master_ip + ":" + master_port
+            rendezvous_iterator = rendezvous(init_method, Utils.rank, Utils.world_size, timeout=timedelta(minutes=1))
+            store, rank, world_size = next(rendezvous_iterator)
```
gshennvm, line 51: nit: rank, world_size seems unused

examples/nlp/gpt/train_gpt_dpo.py
```diff
@@ -138,7 +138,7 @@ def main(cfg) -> None:
 
     logger.log_hyperparams(OmegaConf.to_container(cfg))
 
-    timer = Timer(cfg.exp_manager.get("max_time_per_run"))
+    timer = Timer(cfg.exp_manager.get("max_time_per_run") if cfg.exp_manager else None)
```
gshennvm, line 141: why do you need to guard this? shouldn't it always exist(or is this because of the CI)

tests/functional/ppo.sh
```diff
@@ -0,0 +1,201 @@
+#!/bin/bash
```
gshennvm, line 1: generally speaking i think we should avoid pushing scripts like this, since it's so slurm specific and we can't maintain it

gshennvm (changes requested):

nemo_aligner/algorithms/ppo.py
```diff
@@ -28,8 +28,10 @@
 from nemo.collections.nlp.data.language_modeling.megatron.data_samplers import MegatronPretrainingRandomSampler
 from nemo.collections.nlp.modules.common.megatron.utils import get_iterator_k_split
 from nemo.utils import logging
+from nemo_aligner.models.nlp.gpt.megatron_gpt_ppo_actor import MegatronGPTActorModel
 from nemo_aligner.utils import parallel_state
 from nemo_aligner.utils.distributed import (
+    ScopedTimer,
     SyncTimer,
```
gshennvm, line 35: seems unused

nemo_aligner/algorithms/ppo.py
```diff
@@ -406,46 +399,37 @@ def compute_rollout_metrics(self, rollout_batch):
...
+        _, rollout_metrics = self._run_inference(self.val_dataloader_builder, consumed_samples=0, is_validation=True)
 
         self.model.finish_inference()
         return rollout_metrics
 
     @torch.no_grad()
     def generate_rollouts(self):
-        timing_metrics = {}
+        with self.timer("prepare_for_inference"):
+            self.model.prepare_for_inference(timer=self.timer)
```
gshennvm, line 410: i don't like passing in the timer here, it creates a coupling that is only there for metrics and not logic. I rather us just log this as prep for inference and assume we know that the steps > 1 are refit and not engine building

nemo_aligner/algorithms/ppo.py
```diff
@@ -406,46 +399,37 @@ def compute_rollout_metrics(self, rollout_batch):
...
 
         ppo_rollout_data, ppo_rollout_metrics = self.generate_ppo_data(rollout_batch)
 
-        self.timer.start("finish_inference")
-        self.model.finish_inference()
-        timing_metrics["finish_inference"] = self.timer.stop_and_get_time("finish_inference")
-
-        timing_metrics.update(timer_metrics)
+        with self.timer("finish_inference"):
+            self.model.finish_inference(timer=self.timer)
```
gshennvm, line 421: same concern as above

nemo_aligner/algorithms/ppo.py
```diff
@@ -510,17 +493,18 @@ def fit(self):
                 critic_train_loop_amount = self.critic_warmup_steps + 1 if self.step == 0 else 1
 
                 for _ in range(critic_train_loop_amount):
-                    self.timer.start("rollout_time")
-                    clear_memory()
-                    ppo_rollout_data, metrics, timer_metrics = self.generate_rollouts()
-                    timing_metrics["rollout_time"] = self.timer.stop_and_get_time("rollout_time")
+                    with self.timer("rollout_time"):
+                        clear_memory()
```
gshennvm, line 497: nit: maybe move the clear memory out of the rollout time

gshennvm (changes requested):

nemo_aligner/algorithms/ppo.py
```diff
@@ -569,9 +554,10 @@ def fit(self):
                 )
 
                 if run_val:
-                    self.timer.start("validation_time")
-                    val_metrics = self.run_validation()
-                    timing_metrics["validation_time"] = self.timer.stop_and_get_time("validation_time")
+                    with self.timer("validation_time"):
+                        val_metrics = self.run_validation()
+                    # TODO(terryk): is it okay to log validation_time one step behind?
```
gshennvm, line 559: oh i see, yes there is an off by 1 issue with val. i think that's fine, val at step 5 means we've completed step 4

nemo_aligner/models/nlp/gpt/megatron_gpt_ppo_actor.py
```diff
@@ -98,6 +99,7 @@ def __init__(self, cfg: DictConfig, trainer: Trainer):
                 use_greedy=self.cfg.ppo.sampling_params.get("use_greedy", False),
                 tokenizer=self.tokenizer,
                 seed=self.cfg.ppo.trt_llm.get("seed", self.cfg.seed),
+                trt_model_dir=self.cfg.ppo.get("trt_model_dir", "/tmp/trt_llm_model"),
```
gshennvm, line 102: it will be nice if you also added this onto the PPO actor yaml file too, just so users know it's possible to configure

terrykong: Oops, I think that was dropped. I'll add it back

tests/functional/ppo.sh
terrykong: This script isn't slurm specific. It runs locally

examples/nlp/gpt/train_gpt_dpo.py
terrykong: So there's occasional race conditions when launching with mpirun + experiment manager. I had encountered issues where several ranks will try to os.makedir and fail because the later one errors b/c it already exists

`cfg.exp_manager is None` is valid, so I added this change everywhere since we should be allowed to run without the experiment manager, but I just needed https://github.com/NVIDIA/NeMo/pull/10651

.gitignore
terrykong: Yea

tests/test_mcore_utilities.py
terrykong: Addressed in f16b4cc4b793107d4e23e639cfea29448fd19236

nemo_aligner/algorithms/ppo.py
terrykong: I was trading off the complexity for making the metrics more granular and differentiating offloading adam states with refit measurements

nemo_aligner/models/nlp/gpt/megatron_gpt_ppo_actor.py
terrykong: Added back in: 73f737180aeb5d4afc72140fc66f33d1544693b0

nemo_aligner/algorithms/ppo.py
terrykong: changed TODO to NOTE in 73f737180aeb5d4afc72140fc66f33d1544693b0

nemo_aligner/algorithms/ppo.py
terrykong: done in 73f737180aeb5d4afc72140fc66f33d1544693b0

nemo_aligner/algorithms/ppo.py
terrykong: rm in 63cb223e051ddd9609ad426082883e4aca7a4100

tests/functional/ppo.sh
terrykong: I agree we shouldn't push user scripts like this, but these are functional tests. They need to exist somewhere and if they existed in github action yaml, they'd be hard to reproduce when there's a failure

examples/nlp/gpt/train_gpt_dpo.py
gshennvm: okay this change is fine with me. but shouldn't nemo just guard it themselves? seems like a serious bug in pretty hot piece of code

nemo_aligner/algorithms/ppo.py
gshennvm: I don't think we should do this because it adds an argument that prepare for inference doesn't really need. if we want to fix this properly we should return timing metrics from prepare for inference. but that seems like a seperate PR

nemo_aligner/models/nlp/gpt/megatron_gpt_ppo_actor.py
terrykong: So I looked everywhere and it doesn't appear to be used

nemo_aligner/algorithms/ppo.py
terrykong: reverted

nemo_aligner/algorithms/ppo.py
terrykong: reverted

.github/workflows/_build_container.yml
```diff
@@ -27,13 +27,13 @@ defaults:
...
   main:
-    runs-on: self-hosted-builder
+    runs-on: self-hosted-azure-builder
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
         with:
           path: ${{ github.run_id }}
-          ref: ${{ inputs.sha }}
+          ref: ${{ inputs.ref }}
```
ko3n1g, line 36: I think all of these file changes can get dropped

Dockerfile
```diff
@@ -37,13 +36,41 @@ git checkout -f $ALIGNER_COMMIT
...
+    if [ ! -z $APEX_TAG ]; then \
+        git fetch origin $APEX_TAG && \
+        git checkout FETCH_HEAD; \
+    fi && \
+    pip install -v --no-build-isolation --disable-pip-version-check --no-cache-dir --config-settings "--build-option=--cpp_ext --cuda_ext --fast_layer_norm --distributed_adam --deprecated_fused_adam" ./
+
+# Git LFS
+RUN curl -s https://packagecloud.io/install/repositories/github/git-lfs/script.deb.sh | bash && \
+    apt-get install git-lfs && \
+    git lfs install
```
ko3n1g, line 60:
```suggestion
    git lfs install && \
    apt-get clean
```

Dockerfile
```diff
@@ -37,13 +36,41 @@ git checkout -f $ALIGNER_COMMIT
...
+ARG TRTLLM_VERSION
+RUN git clone https://github.com/NVIDIA/TensorRT-LLM.git && \
+    cd TensorRT-LLM && \
+    git checkout ${TRTLLM_VERSION} && \
+    . docker/common/install_tensorrt.sh && \
+    python3 ./scripts/build_wheel.py --job_count $(nproc) --trt_root /usr/local/tensorrt  --python_bindings --benchmarks
+
+RUN cd TensorRT-LLM && \
+    pip install -e .
+ENV LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/cuda-12/compat/lib.real/
```
ko3n1g, line 72: From caching perspective I don't think it's necessary/helpful to separate this into two stages. WDYT about combining it? 

```suggestion
ARG TRTLLM_VERSION
RUN git clone https://github.com/NVIDIA/TensorRT-LLM.git && \
    cd TensorRT-LLM && \
    git checkout ${TRTLLM_VERSION} && \
    . docker/common/install_tensorrt.sh && \
    python3 ./scripts/build_wheel.py --job_count $(nproc) --trt_root /usr/local/tensorrt  --python_bindings --benchmarks && \
    pip install -e .
ENV LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/cuda-12/compat/lib.real/
```

ko3n1g (changes requested):
very tiny nits, but overall looks great. Hope we can remove the cherry-picking layer soon

nemo_aligner/utils/trt_llm.py
```diff
@@ -48,10 +46,9 @@ def __init__(
         repetition_penalty=1.0,
         max_generation_length=1024,
         max_input_len=1024,
-        max_input_tokens=4096,
         generation_batch_size=4,
         use_greedy=False,
-        trt_model_type="GPTForCausalLM",
+        trt_model_type="llama",
```
gshennvm, line 51: is it safe to change the trt-model-type like this?

terrykong: the GPTForCausalLM was an artifact of the old naming, now we have these [gpt/llama aliases](https://github.com/NVIDIA/NeMo/blob/9d857baf82ce163801459fa7cec71c1e38016b02/nemo/export/trt_llm/converter/utils.py#L30) for those trtllm classes now. actually the [constructor caller](https://github.com/NVIDIA/NeMo-Aligner/blob/main/nemo_aligner/models/nlp/gpt/megatron_gpt_ppo_actor.py#L91) already defaulted properly so this is just to bring this up to date

.github/workflows/_run_test.yml
terrykong: addressed in https://github.com/NVIDIA/NeMo-Aligner/pull/356

.github/workflows/_run_test.yml
terrykong: addressed in https://github.com/NVIDIA/NeMo-Aligner/pull/356

.github/workflows/cicd-main.yml
terrykong: addressed in https://github.com/NVIDIA/NeMo-Aligner/pull/356

ko3n1g (approved):

tests/test_trt_llm.py
```diff
@@ -0,0 +1,84 @@
...
+    # TRTLLM with PP has sometimes erroneously inserts padding:
+    # As an example when we have the input:
+    #     [[prompt tok, PAD, PAD], [prompt tok, prompt tok, prompt tok]]
+    # The output when PP is enabled becomes:
+    #     [[prompt tok, PAD, PAD, resp_tok, resp_tok], [prompt tok, prompt tok, prompt tok, resp_tok, resp_tok]]
+    # Therefore we need this logic to get rid of the padding in the middle of the tensor.
+    # Furthermore, TRTLLM only produces valid outputs on the source rank, so we can only process it here
+    # and rely on the aligner broadcast to get it to the other ranks. Miraculously, the length
+    # is still correct on the non src ranks
+    if (
```
gshennvm, line 49: do we still need this given we've removed it from the main trt-llm file?

terrykong: So my reasons for keeping it are mainly:
1. I don't know what PR fixed this or if this is input dependent, so I thought to leave it for posterity
2. this unit test is also an exemplar of how to test TRTLLM stuff with mpi

gshennvm (approved):

jiemingz (approved):
lgtm!

PR #331, opened 2024-10-01
build: Build trtllm with multi-stage

ko3n1g:
# What does this PR do ?

# Changelog 

# Usage

```python
```

**Pre checks**:

# Additional Information
* Related to # (issue)


Dockerfile.test
terrykong, line 1: Just leaving TODO to remove before merging

terrykong (changes requested):

.github/workflows/_build_container.yml
```diff
@@ -60,19 +70,7 @@ jobs:
...
-      - name: Update buildcache image
-        if: github.ref == 'refs/heads/main'
-        run: |
-          docker tag nemoci.azurecr.io/nemo_aligner_container:${{ github.run_id }} nemoci.azurecr.io/nemo_aligner_container:buildcache
-          docker push nemoci.azurecr.io/nemo_aligner_container:buildcache
\ No newline at end of file
+            nemoci.azurecr.io/nemo_aligner_container_cache:${{ github.event.pull_request.number || 'noop' }}
+            nemoci.azurecr.io/nemo_aligner_container_cache:main
+          cache-to: ${{ env.CACHE }}
+          tags: nemoci.azurecr.io/nemo_aligner_container:${{ github.run_id }}
```
terrykong, line 76: Does docker build always pull the cache image even if the dockerfile is identical? If so, is it possible to speed up the docker build by doing something like this (pseudocode):
```sh
if [[ ! image_tag_exists "nemoci.azurecr.io/nemo_aligner_container:PR-${pr}-$(md5sum Dockerfile) ]] || [[ $CACHEBUST ]]; then
  # do build
else
  # noop
fi
```

ko3n1g: Interesting thought. If we were to use `sed` instead of `build-arg`, this might work. But we'd also be working around Docker's native caching mechanism, and bugs around caching can unfold in an extremely unpleasant way - I'm not sure if we want to go that route yet?

ko3n1g: And then both drivers have the shared property of requiring all layers to be present on the host system for building an updated image (since overlay-fs layers are chained by the merge-dir to their parent)

.github/workflows/_build_container.yml
```diff
@@ -14,20 +14,14 @@
...
   run:
     shell: bash -x -e -u -o pipefail {0}
 
 jobs:
   main:
-    runs-on: self-hosted-builder
+    runs-on: self-hosted-azure-builder
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
```
terrykong, line 27: Should the line three below this be:
```
ref: ${{ inputs.ref }}
```

Dockerfile
```diff
@@ -55,7 +69,21 @@ RUN pip uninstall -y transformer-engine && \
...
+    NVTE_FRAMEWORK=pytorch NVTE_WITH_USERBUFFERS=1 MPI_HOME=/usr/local/mpi pip wheel .
+
+FROM ${TRT_PROXY_IMAGE} as trt-llm-proxy
+FROM ${TE_PROXY_IMAGE} as te-proxy
+
+# Final image
+FROM ${BASE_IMAGE} AS final
+WORKDIR /opt
+# needed in case git complains that it can't detect a valid email, this email is fake but works
+RUN git config --global user.email "worker@nvidia.com"
```
terrykong, line 81: I think since we're fixing things, we should probably unset this at the very end so that folks don't get surprised their committing code as `worker@nvidia.com`

ko3n1g:
Not required anymore

PR #352, opened 2024-10-22
fix: correct batch tokenization when sequence exceeds encoder length

gwarmstrong:
# What does this PR do ?

Fixes a bug in `tokenize_batch` that occurs when there are more tokens than the specified max sequence length.

# Changelog 

# Usage

```python
```

**Pre checks**:

# Additional Information
* Related to # (issue)


tests/test_utils.py
```diff
@@ -54,3 +55,57 @@ def test_select_log_probs():
     # hand derived solution
     assert log_probs[0] == [1, 7], f"expected log probs at idx 0 to be [1,7] but got {log_probs[0]}"
     assert log_probs[1] == [16, 22], f"expected log probs at idx 1 to be [16,22] but got {log_probs[1]}"
+
+
+class _MockTokenizer:
```
terrykong, line 60: No need to protect this class

tests/test_utils.py
terrykong, line 1: I think there are a lot of utils so it would probably help to keep them separated. Could you add your changes to a new `test_text_generation_utils.py`?

tests/test_utils.py
```diff
@@ -54,3 +55,57 @@ def test_select_log_probs():
...
+    sentences = ["I went to the store.", "I bought a zoo."]
+    tokenizer = _MockTokenizer()
+    max_len = 10
+    context_tokens_tensor, context_length_tensor = tokenize_batch(
+        sentences, tokenizer, max_len, add_BOS=False, add_EOS=False
+    )
+    assert context_tokens_tensor.shape == (
+        2,
+        10,
+    ), f"expected context_tokens_tensor shape to be (2, 30) but got {context_tokens_tensor.shape}"
```
terrykong, line 104: comment doesn't seem to match assertion

terrykong (changes requested):
Great catch! I've left my comments below

tests/test_utils.py
gwarmstrong: fixed

tests/test_utils.py
gwarmstrong: added

tests/test_utils.py
gwarmstrong: made it public

terrykong:
Needs #354

terrykong (approved):

PR #361, opened 2024-10-25
feat: update reward model to support scaled and margin BT

Zhilin123:
# What does this PR do ?

Update reward model to support scaled and margin BT

# Changelog 

# Usage

```python
```

**Pre checks**:
- [x] Make sure you read and followed [Contributor guidelines](/CONTRIBUTING.md)
- [x] Did you write any new necessary tests?

# Additional Information
* Related to # (issue)


nemo_aligner/data/nlp/builders.py
```diff
@@ -341,7 +341,7 @@ def build_dataloader(
     pad_samples_to_global_batch_size=False,
     collate_fn=None,
     load_gbs=True,
-    use_random_sampler=True,
+    use_random_sampler=False,
```
HeyyyyyyG, line 344: Is there a reason to set default as False?

nemo_aligner/models/nlp/gpt/megatron_gpt_regression_reward_model.py
```diff
@@ -114,6 +114,47 @@ def loss_func(output_tensor):
...
 
     def loss_func(self, output_tensor, label_tensor):
+        loss_func_name = self.cfg.get("loss_func", "regression")
+
+        if loss_func_name == "regression":
+            return self.reg_loss_func(output_tensor, label_tensor)
+        elif loss_func_name.endswith("bt"):
+            return self.bt_loss_func(output_tensor, label_tensor)
+        else:
+            raise ValueError("only accepted values for loss_func are regression, regular_bt, margin_bt and scaled_bt")
```
HeyyyyyyG, line 124: maybe add a check for these types explicitly? something like 
`assert loss_func_name in ["regression", "regular_bt", "margin_bt" and "scaled_bt"]`

nemo_aligner/models/nlp/gpt/megatron_gpt_regression_reward_model.py
```diff
@@ -114,6 +114,47 @@ def loss_func(output_tensor):
...
+
+    def bt_loss_func(self, output_tensor, label_tensor):
+        """
+        label_tensor should have the value of the [0,0] index as a number indicating the prefered response as well as the strength of the preferred response
+        
+        In HelpSteer2-Preference, label is a non-negative integer between -3 and 3: -3, -2, -1 means A is preferred while 1, 2, 3 means B is preferred
+        
+        output_tensor uses the value of [:, 4] index as the reward for the chosen/rejected values, which inherits the position of the helpfulness attribute
+        """
+        label = label_tensor[0, 0]
```
HeyyyyyyG, line 134: I'm a bit confused here. Could you give an example of label_tensor, its shape etc?

examples/nlp/gpt/conf/training_rm.yaml
```diff
@@ -63,10 +63,12 @@ model:
     merge_attributes: False # whether to merge multiple attributes into a scalar
     attribute_weights: null # apply these weights to each attributes when merging them into a scalar
     loss_mask_val: -100 #  mask dimensions with this value when calculating MSE loss
+    loss_func: regression # ["regression", "regular_bt", "margin_bt", "scaled_bt"]
+    load_rm_head_weights: False # [False, True] False only loads base model while True loads rm_head weights as well (useful for intializing rm_head with model containing existing rm_head)
   output_sequence: False  # Whether to output a single scalar or a sequence of scalars.
   use_avg_pool: False  # Whether to use avg pool to sum across the sequence dim in reward model
   force_head_dtype: float32  # enforce specific dtype for the final projection in the model head
-  micro_batch_size: 1
+  micro_batch_size: 2 # please do not adjust MBS to other values for xxx_bt implementations
```
HeyyyyyyG, line 71: Better to add a check somewhere

nemo_aligner/data/nlp/builders.py
Zhilin123: Yes, if this is not set to False, then it will jumble all of the samples up and two responses to the same prompt will not appear in the same microbatch, which is needed for xxx_BT

nemo_aligner/models/nlp/gpt/megatron_gpt_regression_reward_model.py
Zhilin123: sure

examples/nlp/gpt/conf/training_rm.yaml
Zhilin123: done - added to ```examples/nlp/gpt/train_reward_model.py```

nemo_aligner/models/nlp/gpt/megatron_gpt_regression_reward_model.py
Zhilin123: done - is this clearer?

examples/nlp/gpt/conf/training_rm.yaml
```diff
@@ -63,10 +63,12 @@ model:
     merge_attributes: False # whether to merge multiple attributes into a scalar
     attribute_weights: null # apply these weights to each attributes when merging them into a scalar
     loss_mask_val: -100 #  mask dimensions with this value when calculating MSE loss
+    loss_func: regression # ["regression", "regular_bt", "margin_bt", "scaled_bt"]
+    load_rm_head_weights: False # [False, True] False only loads base model while True loads rm_head weights as well (useful for intializing rm_head with model containing existing rm_head)
   output_sequence: False  # Whether to output a single scalar or a sequence of scalars.
   use_avg_pool: False  # Whether to use avg pool to sum across the sequence dim in reward model
-  force_head_dtype: float32  # enforce specific dtype for the final projection in the model head
-  micro_batch_size: 1
+  force_head_dtype: bfloat16 #float32  # enforce specific dtype for the final projection in the model head
```
Zhilin123, line 70: @HeyyyyyyG I want to give you a heads up that this has been changed to bfloat16

The main reason is the CI doesn't pass with float32 (likely due to newer mcore version). Also float32 for just the reward head (everywhere else is bfloat16) isn't supported by TRTLLM or HF compatible checkpoint. My test of post-training-converted bfloat16 weights shows little degradation so I think it's ok to keep it as bfloat16.

```
[rank1]:   File "/opt/NeMo-Aligner/nemo_aligner/algorithms/supervised.py", line 210, in fit
[rank1]:     loss, metrics = self.train_single_step(batch)
[rank1]:   File "/opt/NeMo-Aligner/nemo_aligner/algorithms/supervised.py", line 145, in train_single_step
[rank1]:     loss_mean, metrics = self.model.get_loss_and_metrics(batch=batch, forward_only=False)
[rank1]:   File "/opt/NeMo-Aligner/nemo_aligner/models/nlp/gpt/megatron_gpt_regression_reward_model.py", line 182, in get_loss_and_metrics
[rank1]:     losses_reduced_per_micro_batch = fwd_bwd_function(
[rank1]:   File "/opt/Megatron-LM/megatron/core/pipeline_parallel/schedules.py", line 474, in forward_backward_no_pipelining
[rank1]:     backward_step(input_tensor, output_tensor, output_tensor_grad, model_type, config)
[rank1]:   File "/opt/Megatron-LM/megatron/core/pipeline_parallel/schedules.py", line 354, in backward_step
[rank1]:     custom_backward(output_tensor[0], output_tensor_grad[0])
[rank1]:   File "/opt/Megatron-LM/megatron/core/pipeline_parallel/schedules.py", line 154, in custom_backward
[rank1]:     Variable._execution_engine.run_backward(
[rank1]:   File "/usr/local/lib/python3.10/dist-packages/torch/autograd/function.py", line 301, in apply
[rank1]:     return user_fn(self, *args)
[rank1]:   File "/usr/local/lib/python3.10/dist-packages/torch/cuda/amp/autocast_mode.py", line 142, in decorate_bwd
[rank1]:     return bwd(*args, **kwargs)
[rank1]:   File "/opt/Megatron-LM/megatron/core/tensor_parallel/layers.py", line 497, in backward
[rank1]:     fused_weight_gradient_mlp_cuda.wgrad_gemm_accum_fp16(
[rank1]: RuntimeError: expected scalar type BFloat16 but found Float
```

Zhilin123: see this CI https://github.com/NVIDIA/NeMo-Aligner/actions/runs/11565227336/job/32191971677#step:5:716

HeyyyyyyG: cc @odelalleau

nemo_aligner/data/nlp/builders.py
HeyyyyyyG: @gshennvm I think this will also affect dpo, ppo etc, is it ok to set as False?

nemo_aligner/models/nlp/gpt/megatron_gpt_regression_reward_model.py
HeyyyyyyG: Yes , thanks

HeyyyyyyG:
@Zhilin123 Can we add a doc to show how to reproduce the RM training process(steps, data preparation etc)?

nemo_aligner/data/nlp/builders.py
gshennvm: this is tricky... for the other algorithms it is technically ok to set it to False because we do "pairs" of dataset rather than single samples. but this changes the default behavior and means users cannot run multiple epochs without changing the code.


how about we push a flag into the yaml that defaults to True, but we can set this to False? something like model.data.train_ds.use_random_sampler

examples/nlp/gpt/conf/training_rm.yaml
gshennvm: but we'd be training with BF16 head and I believe there was an issue with this before

ashors1:
> @Zhilin123 Can we add a doc to show how to reproduce the RM training process(steps, data preparation etc)?

+1, some documentation on this would be super useful

examples/nlp/gpt/conf/training_rm.yaml
Zhilin123: @gshennvm What was the issue?

nemo_aligner/data/nlp/builders.py
Zhilin123: Why does this affect multiple epoch training? 

In general, I think perhaps the easier approach is to simply have it as default shuffle = False across all algorithms. If user wants to shuf the training set, they can do it by themselves (e.g. with shuf on bash beforehand during data preparation). This reduces replication issues due to differences in shuffling on different settings (this way for shuffling was only introduced in 24.05 release AND in the future, we cannot guarantee that it will be done in the same way).

Also, right now, there is no way to disable shuffling across all algorithms to the best of my knowledge.

gshennvm: yeah right now it's not great. we want to enable random shuffle because if the user naively trains for many epochs without random shuffle then the next epoch will repeat in the same order if we don't use this random shuffle

examples/nlp/gpt/conf/training_rm.yaml
gshennvm: I believe it was convergence but @odelalleau ran those experiments so best for him to give details there

nemo_aligner/data/nlp/builders.py
Zhilin123: ```if the user naively trains for many epochs without random shuffle then the next epoch will repeat in the same order ``` --> this is actually the expected behavior [if each sample is represented by a letter i.e. DACB DACB DACB  is the right order rather than DACB CADB BACD ] The rationale is that If I trained for 1.5 epochs, and the job ended and then I resume training, I want to be sure that the next 0.5 epochs is something that was not seen yet in the second epoch) . Why is it bad to sample in the same order?

Zhilin123:
> > @Zhilin123 Can we add a doc to show how to reproduce the RM training process(steps, data preparation etc)?
> 
> +1, some documentation on this would be super useful

@HeyyyyyyG / @ashors1 - I prepared `NeMo-Aligner/examples/nlp/data/steerlm/preprocess_helpsteer2_data.py`, which should be a drop in replacement for `/home/zhilinw/NeMo-Aligner/examples/nlp/data/steerlm/preprocess_helpsteer_data.py` based on this doc https://docs.nvidia.com/nemo-framework/user-guide/latest/modelalignment/steerlm.html#step-2-download-and-preprocess-data-for-attribute-prediction-modelling

Let me know if you think more documentation is needed.

nemo_aligner/data/nlp/builders.py
HeyyyyyyG: ` this is actually the expected behavior` -> I don't think people have aligned on this. Random order for each epoch may help alleviate overfitting in some cases. (e.g. I heard the long context team wants this behaviour)

HeyyyyyyG:
> > > @Zhilin123 Can we add a doc to show how to reproduce the RM training process(steps, data preparation etc)?
> > 
> > 
> > +1, some documentation on this would be super useful
> 
> @HeyyyyyyG / @ashors1 - I prepared `NeMo-Aligner/examples/nlp/data/steerlm/preprocess_helpsteer2_data.py`, which should be a drop in replacement for `NeMo-Aligner/examples/nlp/data/steerlm/preprocess_helpsteer_data.py` based on this doc https://docs.nvidia.com/nemo-framework/user-guide/latest/modelalignment/steerlm.html#step-2-download-and-preprocess-data-for-attribute-prediction-modelling
> 
> Let me know if you think more documentation is needed.

I believe the recipe is to first train a regression model then finetune it with bt loss. Do we have an example code for this documented somewhere?

examples/nlp/gpt/conf/training_rm.yaml
Zhilin123: After discussion with @gshennvm and @odelalleau , we decided to keep it as BF16 by default.

nemo_aligner/data/nlp/builders.py
Zhilin123: After discussions with @odelalleau and @gshennvm , we will not change the default behavior but make the shuffling behavior of both train and val user configurable.

Zhilin123:
> > > > @Zhilin123 Can we add a doc to show how to reproduce the RM training process(steps, data preparation etc)?
> > > 
> > > 
> > > +1, some documentation on this would be super useful
> > 
> > 
> > @HeyyyyyyG / @ashors1 - I prepared `NeMo-Aligner/examples/nlp/data/steerlm/preprocess_helpsteer2_data.py`, which should be a drop in replacement for `NeMo-Aligner/examples/nlp/data/steerlm/preprocess_helpsteer_data.py` based on this doc https://docs.nvidia.com/nemo-framework/user-guide/latest/modelalignment/steerlm.html#step-2-download-and-preprocess-data-for-attribute-prediction-modelling
> > Let me know if you think more documentation is needed.
> 
> Thanks. I believe the recipe is to first train a regression model then finetune it with bt loss. Do we have an example code for this documented somewhere?

I see - I updated the SteerLM tutorial to reflect this - see changes to ```docs/user-guide/steerlm.rst```

docs/user-guide/steerlm.rst
```diff
@@ -76,7 +76,13 @@ Untar the .nemo file to obtain the tokenizer in NeMo format (only for the 70B mo
 
 The prefix for the tokenizer would be different when extracted. Ensure that the correct tokenizer file is used when running the preceding command.
 
-Step 2: Download and Preprocess data for Attribute Prediction Modelling
+To follow the HelpSteer2 and HelpSteer2-Preference line of works, you would need to use the Llama 3 70B and Llama 3.1 70B Instruct models respectively.
+
+You need to get access to them, download and then convert them in a similar manner.
+
+An additional requirement is that you need to have an HF_TOKEN with access to the relevant models and then either export HF_TOKEN=<your_hf_token> or HF_HOME=<your_hf_home> contain a file named token with your token.
```
ashors1, line 83: This line can be removed. We're adding some documentation on `HF_TOKEN`s that will apply to all models

docs/user-guide/steerlm.rst
```diff
@@ -108,6 +114,53 @@ Finally, preprocess the data into regression reward model training format.
       --output-file=data/merge_val_reg.jsonl
 
 
+If you are interested in following the approach from HelpSteer2 work, please do the below instead.
```
ashors1, line 117: Can you provide a bit more detail on what it means to "follow the approach"? A brief explanation of the approach and how it differs from the standard steerlm approach would be useful.

docs/user-guide/steerlm.rst
Zhilin123: sure done

docs/user-guide/steerlm.rst
Zhilin123: The approach is the exactly the same as standard SteerLM approach. I changed the wording to "If you are interested in replicating Reward Modelling training in HelpSteer2, please do the below instead."

HeyyyyyyG:
Thanks for the changes, LGTM!

tests/functional/rm.sh
```diff
@@ -0,0 +1,84 @@
...
+CONF_NAME="training_rm"
+
+CHECKPOINT_DIR="${RESULTS_DIR}/checkpoints"
+TENSOBOARD_DIR="${RESULTS_DIR}/tensorboard"
+
+mkdir -p $RESULTS_DIR
+mkdir -p $TENSOBOARD_DIR
+mkdir -p $CHECKPOINT_DIR
+
+dpo() {
```
HeyyyyyyG, line 43: minor typo, should be `rm`

tests/functional/rm.sh
```diff
@@ -0,0 +1,84 @@
...
+    ++model.activations_checkpoint_method="uniform" \
+    ++model.activations_checkpoint_num_layers=1 \
+    ++model.sequence_parallel=False \
+    model.reward_model_type="regression" \
+    ++model.dist_ckpt_load_strictness=log_all \
+    model.regression.num_attributes=9
+}
+
+log_file=$(mktemp /tmp/rm-log-XXXXXX)
+dpo | tee $log_file
```
HeyyyyyyG, line 84: minor typo, should be `rm`

docs/user-guide/steerlm.rst
```diff
@@ -106,8 +111,56 @@ Download and Preprocess Data for Attribute Prediction Modeling
          --output-file=data/merge_val_reg.jsonl
 
 
-Train the Regression Reward Model on OASST+HelpSteer Data
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+If you are interested in replicating Reward Modelling training in HelpSteer2, please do the below instead.
```
jgerh, line 114: Line 43 fix capitalization (use title case for headings).

Train a SteerLM Model 


Line 51/51 fix capitalization  (use title case for headings).

Download the Llama 2 LLM Model 


Line 52/52 add ^ characters to align with length of heading.

Download the Llama 2 LLM Model 
^^^^^^^^^^^^^^^^^^^^^^^^^^


Line 78 revise sentence, fix punctuation.

To follow the HelpSteer2 and HelpSteer2-Preference line of works, you need to use the LLama 3 70B and LLama 3.1 70B Instruct models, respectively.


Line 80 revise sentence.

You need to obtain access to them, download them, and then convert them in a similar manner.


Line 82 remove numbered step, fix capitalization, fix spelling, and change heading to a sub-heading format (use ^ characters).

Download and Preprocess Data for SteerLM Regression Reward Modeling
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^


Line 88/93 fix punctuation.

#. Merge the two datasets for the train and val subset, respectively:


Line 114 revise sentence, fix spelling.

If you are interested in replicating Reward Modeling training in HelpSteer2, please follow the steps below instead.


Line 131 revise sentence, fix spelling.

If you're interested in replicating Reward Modeling training in HelpSteer2-Preference, please follow the steps below instead.


Line 161 remove numbered step, fix capitalization, fix spelling, and change heading to a sub-heading format (use ^ characters).

Train the Regression Reward Model on OASST+HelpSteer Data
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^


Line 112/165 revise sentence.

In this tutorial, you train the regression reward model for 800 steps. 


Line 194 revise sentence, fix spelling.

If you're interested in replicating Reward Modeling training in HelpSteer2, please follow the steps below instead.


Line 226 revise sentence, fix spelling.

If you're interested in replicating Reward Modeling training in HelpSteer2-Preference, please follow the steps below instead.

tests/functional/rm.sh
Zhilin123: ok I'll call it `rm_training` to avoid accidentally overlapping with `rm` as in the linux bash operation

docs/user-guide/steerlm.rst
Zhilin123: fixed - thanks!

ashors1 (approved):

terrykong (changes requested):
@Zhilin123 Thanks for adding this feature!

Could you please update the CHANGELOG.md with a line describing this feature so that we may highlight it and users know what was added?

Zhilin123:
> @Zhilin123 Thanks for adding this feature!
> 
> Could you please update the CHANGELOG.md with a line describing this feature so that we may highlight it and users know what was added?

Done!

CHANGELOG.md
```diff
@@ -34,6 +34,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
     # Consume all durations and reset internal store
     durations = timer.consume_durations()
     ```
+- Add code and instructions for replicating Reward Modelling training in HelpSteer2 and HelpSteer2-Preference
```
terrykong, line 37:
```suggestion
- Add code and instructions for replicating Reward Modeling training in HelpSteer2 and HelpSteer2-Preference
```

terrykong (approved):

terrykong (approved):

PR #380, opened 2024-11-05
feat: support for mcore optimizer (to enable MoE)

terrykong:
# What does this PR do ?
- Finishes up support for mcore optimizer
- adds tests for mixtral EP and mixtral TP + SP + peft
    - EP is not fully functional in this combination of nemo/mcore. disabling test for now

## Rebase stack
* https://github.com/NVIDIA/NeMo-Aligner/pull/386
* https://github.com/NVIDIA/NeMo-Aligner/pull/387
* https://github.com/NVIDIA/NeMo-Aligner/pull/388
* https://github.com/NVIDIA/NeMo-Aligner/pull/380

# Changelog 

# Usage

```python
```

**Pre checks**:

# Additional Information
* Related to # (issue)


terrykong:
The DPO dataset changes should stand on their own, but are needed to test the mcore opt changes for moe. If moe issues take too long to resolve, I'll break this up.

ashors1 (approved):

nemo_aligner/utils/train_utils.py
```diff
@@ -101,31 +101,52 @@ def prepare_for_training_step(ptl_model, zero_grad=True):
...
     if ptl_model.cfg.get("tensor_model_parallel_size", 1) > 1 and ptl_model.cfg.get("sequence_parallel", False):
-        ptl_model.allreduce_sequence_parallel_gradients()
-
-    if ptl_model.with_distributed_adam:
-        # synchronize asynchronous grad reductions
-        # note: not necessary, but reduces performance degradation
-        # from multiple simultaneous NCCL calls
-        ptl_model._optimizer._finish_bucket_grad_sync()
+        # Mcore DistOpt handles this, so we don't have to
+        if not ptl_model.use_mcore_dist_optim:
```
gshennvm, line 110: how do you feel about dropping support for non mcore dist optim? are they equivalent to apex now?

nemo_aligner/utils/train_utils.py
```diff
@@ -101,31 +101,52 @@ def prepare_for_training_step(ptl_model, zero_grad=True):
...
+            or ptl_model.cfg.get("sequence_parallel", False)
+            or not ptl_model.cfg.get("async_grad_allreduce", True)
+        ):
             # main grads are stored in the MainParamsOptimizer wrapper
             ptl_model._optimizer.allreduce_main_grads()
     else:
         # async grad allreduce is not currently implemented for O1/autocasting mixed precision training
         # so we all-reduce gradients after the pipeline
         ptl_model.allreduce_gradients()  # @sangkug we think this is causing memory to blow up (hurts perf)
+    ptl_model.megatron_timer_stop("gradient_allreduce")
```
gshennvm, line 139: where is the start here?(maybe i just missed it?)

nemo_aligner/utils/train_utils.py
terrykong: yea, i want to do that in a follow up PR (would help our build times immensely). This just adds the feature without breaking apex

nemo_aligner/utils/train_utils.py
terrykong: https://github.com/NVIDIA/NeMo-Aligner/pull/380/files/4f60e8729a1f7b3683ab92bfa4b1f735aa16f150#diff-2fe1ae18d9ee9628d1cc64471a6124b6bb7c8c8cd9e2a76829d125393a83fd0aR115

nemo_aligner/algorithms/supervised.py
```diff
@@ -150,7 +150,7 @@ def train_single_step(self, batch):
         grad_norm = grad_norm.item() if torch.is_tensor(grad_norm) else grad_norm
         lr = self.optimizer.param_groups[0]["lr"]
 
-        self.optimizer.step()
+        self.optimizer.step(closure=None)
```
gshennvm, line 153: do we have to specify the closure now? i thought this was optional?

terrykong: for mcore dist opt it's required, so i just set it everywhere

examples/nlp/gpt/conf/training_rm.yaml
```diff
@@ -6,6 +6,7 @@ trainer:
   devices: 8
   accelerator: gpu
   precision: bf16
+  gradient_clip_val: 0.0  # No need to change. Megatron Core optimizer uses this value
```
gshennvm, line 9: is mcore picking up this value properly? if that's the case we should default it to the previous 1.0 to avoid any surprises with the switch

gshennvm: if the above is the case, maybe it's better to pin the value as `${trainer.rm.gradient_clip_val}`

gshennvm (changes requested):

examples/nlp/gpt/conf/training_rm.yaml
ashors1: If we set the default to 1.0, I believe we end up doing gradient clipping twice (once in mcore, once in aligner). @terrykong please correct me if I'm wrong here

ashors1: Oop never mind, the later bits of the code indicate that we expect clipping to be done from mcore

nemo_aligner/utils/train_utils.py
```diff
@@ -101,31 +101,52 @@ def prepare_for_training_step(ptl_model, zero_grad=True):
                     param.data_ptr()
 
 
+# TODO: Delete this once API introduced in NeMo (https://github.com/NVIDIA/NeMo/pull/10803)
+# TODO: Update PR to move this logic into staticmethod in nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py
 def grad_reductions(ptl_model):
```
ashors1, line 106: Is there a reason we can't leverage mcore's [finalize_model_grads](https://github.com/NVIDIA/Megatron-LM/blob/c7bf403ba37ce7ca739a025accac232c75a38966/megatron/core/distributed/finalize_model_grads.py#L212) function directly?

terrykong: I think we could try to use that instead, but I'd like to defer any cleanup to our refactoring effort. My goal in this PR is to at least reach parity to nemo 1.0 so that we can remove apex ahead of the refactoring effort. WDYT?

examples/nlp/gpt/conf/training_rm.yaml
terrykong: > if the above is the case, maybe it's better to pin the value as ${trainer.rm.gradient_clip_val} 

Good point. It is probably more jarring to switch over and not get gradient clipping (if it was 0.0). I applied that suggestion in https://github.com/NVIDIA/NeMo-Aligner/pull/380/commits/149f4aaad160d9f4fe5be8589a5200e3cc48b690

nemo_aligner/utils/train_utils.py
ashors1: That makes sense to me, thanks!

ashors1 (approved):

gshennvm (approved):

PR #405, opened 2024-11-14
feat: support new DPO data format and update SFT config to use override API

arendu:
# What does this PR do ?
This PR makes the dpo dataset use chat format tokens from the model's config yaml instead of hardcoding chat/special tokens in the jsonl data file.

Currently, each datapoint inside a DPO jsonl data file, looks like this:
```
{
  "prompt": "<extra_id_0>System\n\n<extra_id_1>User\nbacillus subtilus\n<extra_id_1>Assistant\n",
  "chosen_response": "Bacillus ... and industry alike.\n<extra_id_1>",
  "rejected_response": "The Bacillus ... fields of study.\n<extra_id_1>",
  "rejected_reward": 3,
  "chosen_reward": 4
}
```
With this PR it should be like this (OpenAI list of messages format with no chat/formatting tokens):
```
{
  "prompt": [
    {
      "role": "system",
      "content": ""
    },
    {
      "role": "user",
      "content": "bacillus subtilus"
    }
  ],
  "chosen_response": {
    "role": "assistant",
    "content": "Bacillus ... and industry alike."
  },
  "rejected_response": {
    "role": "assistant",
    "content": "The Bacillus ... fields of study."
  },
  "chosen_reward": 4,
  "rejected_reward": 3
}
```

*Additionally* There is a script added to convert old data files into the new format.
```bash
python nemo_aligner/data/nlp/scripts/undo_special_tokens.py <path_to_old_format_dpo_jsonl_file>
```
A new file will be written in the same location as the old format file.

# Changelog 

# Usage

```python
```

**Pre checks**:

# Additional Information
* Related to # (issue)

### DIFF
--- a/examples/nlp/gpt/train_gpt_sft.py
+++ b/examples/nlp/gpt/train_gpt_sft.py
@@ -39,7 +39,7 @@ from nemo_aligner.utils.train_script_utils import (
     resolve_and_create_trainer,
     retrieve_custom_trainer_state_dict,
 )
-from nemo_aligner.utils.utils import load_from_nemo
+from nemo_aligner.utils.utils import load_and_override_model_config, load_from_nemo
 
 """Script to start SFT training"""
 
@@ -49,75 +49,10 @@ OmegaConf.register_new_resolver("int_div", lambda x, y: x // y, replace=True)
 mp.set_start_method("spawn", force=True)
 
 
-def _modify_config(gpt_cfg, cfg, add_cfg_to_tree=False):
-    """
-    This function modifies the original gpt pre-training config (gpt_cfg) with attributes from the finetuning config (cfg).
-    The `add_cfg_to_tree` arg adds `cfg` to the top of the yaml tree which is needed for all `hparams.yaml` files when passed as an arg to `load_from_checkpoint()`.
-    """
-    OmegaConf.set_struct(gpt_cfg, True)
-    OmegaConf.resolve(cfg)
-    with open_dict(gpt_cfg):
-        gpt_cfg.megatron_amp_O2 = cfg.model.get("megatron_amp_O2", False)
-        gpt_cfg.micro_batch_size = cfg.model.data.train_ds.micro_batch_size
-        gpt_cfg.global_batch_size = cfg.model.data.train_ds.global_batch_size
-        gpt_cfg.sequence_parallel = cfg.model.get("sequence_parallel", False)
-        gpt_cfg.activations_checkpoint_granularity = cfg.model.get("activations_checkpoint_granularity", None)
-        gpt_cfg.activations_checkpoint_num_layers = cfg.model.get("activations_checkpoint_num_layers", None)
-        gpt_cfg.activations_checkpoint_method = cfg.model.get("activations_checkpoint_method", None)
-        gpt_cfg.activations_checkpoint_layers_per_pipeline = cfg.model.get(
-            "activations_checkpoint_layers_per_pipeline", None
-        )
-        gpt_cfg.peft = cfg.model.peft
-        gpt_cfg.data = cfg.model.data
-        gpt_cfg.optim = cfg.model.optim
-        gpt_cfg.precision = cfg.trainer.precision
-        gpt_cfg.answer_only_loss = cfg.model.answer_only_loss
-        gpt_cfg.restore_from_path = cfg.model.restore_from_path
-        gpt_cfg.resume_from_checkpoint = cfg.model.resume_from_checkpoint
-        gpt_cfg.save_nemo_on_validation_end = cfg.model.save_nemo_on_validation_end
-        gpt_cfg.gradient_as_bucket_view = cfg.model.gradient_as_bucket_view
-        gpt_cfg.hidden_dropout = cfg.model.get("hidden_dropout", 0.0)
-        gpt_cfg.attention_dropout = cfg.model.get("attention_dropout", 0.0)
-        gpt_cfg.ffn_dropout = cfg.model.ffn_dropout
-        gpt_cfg.use_flash_attention = cfg.model.get("use_flash_attention", False)
-        # if TP/PP size is -1, use default TP/PP size as original model
-        if cfg.model.get("tensor_model_parallel_size", 1) > 0:
-            gpt_cfg.tensor_model_parallel_size = cfg.model.get("tensor_model_parallel_size", 1)
-        if cfg.model.get("pipeline_model_parallel_size", 1) > 0:
-            gpt_cfg.pipeline_model_parallel_size = cfg.model.get("pipeline_model_parallel_size", 1)
-        gpt_cfg.pipeline_model_parallel_split_rank = cfg.model.get("pipeline_model_parallel_split_rank", 0)
-
-        if cfg.model.data.get("chat", False):
-            # chat model, overwrite the prompt template
-            prompt_template = get_prompt_template_example(cfg.model.data.chat_prompt_tokens)
-            gpt_cfg.data.train_ds.prompt_template = prompt_template
-            gpt_cfg.data.validation_ds.prompt_template = prompt_template
-
-        sft_cls = GPTSFTModel
-        gpt_cfg.target = f"{sft_cls.__module__}.{sft_cls.__name__}"
-
-        if cfg.model.get("use_flash_attention", None) is not None:
-            gpt_cfg.use_flash_attention = cfg.model.use_flash_attention
-
-        if cfg.model.get("seq_len_interpolation_factor", None) is not None:
-            gpt_cfg.seq_len_interpolation_factor = cfg.model.seq_len_interpolation_factor
-
-        if cfg.model.get("dist_ckpt_load_strictness", None) is not None:
-            gpt_cfg.dist_ckpt_load_strictness = cfg.model.dist_ckpt_load_strictness
-
-        gpt_cfg.inference = cfg.model.get("inference", {})
-
-        # This is needed when modifying a hparam file directly to load `.ckpt` files.
-        # This is not needed to modify the cfg in `.nemo` files.
-        if add_cfg_to_tree:
-            OmegaConf.resolve(gpt_cfg)
-            gpt_cfg.cfg = gpt_cfg
-
-    return gpt_cfg
-
-
 @hydra_runner(config_path="conf", config_name="gpt_sft")
 def main(cfg) -> None:
+    cfg.model = load_and_override_model_config(cfg.model.restore_from_path, cfg.model)
+
     logging.info("\n\n************** Experiment configuration ***********")
     logging.info(f"\n{OmegaConf.to_yaml(cfg)}")
 
@@ -129,17 +64,11 @@ def main(cfg) -> None:
     with open_dict(cfg):
         cfg.model.precision = cfg.trainer.precision
 
-    ptl_model, updated_cfg = load_from_nemo(
-        GPTSFTModel,
-        cfg,
-        trainer,
-        strict=True,
-        modify_config_fn=_modify_config,
-        restore_path=cfg.model.restore_from_path,
-        return_updated_cfg=True,
+    ptl_model = load_from_nemo(
+        GPTSFTModel, cfg, trainer, strict=True, restore_path=cfg.model.restore_from_path, return_updated_cfg=False,
     )
 
-    init_peft(ptl_model, updated_cfg)
+    init_peft(ptl_model, cfg.model)
 
     with open_dict(cfg):
         # overwrite the model config with the config from the checkpoint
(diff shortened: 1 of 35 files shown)


nemo_aligner/data/nlp/datasets.py
terrykong, line 1: could you add some static typing for at least the new methods you are adding? Over time we should converge on the whole repo being staticly typed

nemo_aligner/data/nlp/scripts/undo_special_tokens.py
```diff
@@ -0,0 +1,75 @@
...
+# limitations under the License.
+
+"""Script to remove special tokens from dpo datasets 
+and convert them into list of messages format"""
+
+import json
+import re
+import sys
+
+input_jsonl = sys.argv[1]
```
terrykong, line 22: could we write this in the:

```python
if __name__ == '__main__':
```

pattern? It helps when someone needs to import modules to test. If you refactor, can you also write a unit test to make sure this function does what you expect so future folks don't break the desired behavior?

nemo_aligner/data/nlp/scripts/undo_special_tokens.py
```diff
@@ -0,0 +1,75 @@
...
...or handling of large input datasets\n -  Not as accurate as transformers\n -  Not as flexible as transformers\n \n\n Pros of transformers:\n \n\n -  Better parallelism and handling of large input datasets\n -  More flexible and versatile\n -  Higher accuracy\n -  Good for actual applications in industry and the real world\n \n\n Cons of transformers:\n \n\n -  More complex and harder to understand and debug\n -  Slower to run\n -  Not as good for teaching and research\n<extra_id_1>User\nFor a chat-based music recommendation platform, which model is recommended? Please mention any model that works well, including those mentioned above or any others.\n<extra_id_1>Assistant\nFor a chat-based music recommendation platform, I would recommend using a transformer architecture, such as the GPT-4 model. Transformers are better for handling large input datasets, and they are more flexible and versatile than RNNs. They are also more accurate, so they are better for actually recommending music to users in a real-world application. RNNs are better for teaching and research, as they are simpler and easier to understand and debug.\n<extra_id_1>User\ncould you add more?\n<extra_id_1>Assistant\n"""
```
terrykong, line 65: i'm guessing this is for debugging, but this section is actually a great candidate for a unit test :)

nemo_aligner/data/nlp/datasets.py
terrykong, line 1: Also, could you write a unit test for compatibility with the prompt/chosen/rejected format and the openai format and ensure they both are equivalent given the correct configuration?

nemo_aligner/data/nlp/datasets.py
```diff
@@ -344,16 +348,63 @@ def encode(self, text, append_eod=False):
 
         return text_ids, len(text_ids)
 
+    def _convert_messages(self, input_list):  # TODO: (@adithyare) this method should live elsewhare..
```
terrykong, line 351: Function seems pure so would be good to decorate with `@staticmethod`. This also means it'd be nice to add a unit test for this.

I'm not sure what this is converting to and from, could you add a docstring or give the function a name informing what we are converting to and from?

nemo_aligner/data/nlp/datasets.py
```diff
@@ -344,16 +348,63 @@ def encode(self, text, append_eod=False):
...
+        # Build the conversations list
+        for msg in input_list:
+            if msg["role"] != "system":
+                conversation_entry = {
+                    "from": msg["role"].capitalize(),  # Capitalize 'user' and 'assistant'
+                    "value": msg["content"],
+                    "label": None,
+                }
+                output_dict["conversations"].append(conversation_entry)
+
```
terrykong, line 374: if the user gets here and the input didn't have the keys you expect, do you think it'd help to add some assertions here to make sure the code doesn't error somewhere else, when the issue was they didn't pass the right input?

examples/nlp/gpt/train_gpt_dpo.py
```diff
@@ -37,6 +37,7 @@
 
 OmegaConf.register_new_resolver("multiply", lambda x, y: x * y, replace=True)
 OmegaConf.register_new_resolver("int_div", lambda x, y: x // y, replace=True)
+OmegaConf.register_new_resolver("not", lambda x: not x, replace=True)
```
terrykong, line 40: same as the variable_seq_lengths change, i didn't end up needing this

terrykong (changes requested):

examples/nlp/gpt/train_gpt_sft.py
```diff
@@ -118,6 +118,8 @@ def _modify_config(gpt_cfg, cfg, add_cfg_to_tree=False):
 
 @hydra_runner(config_path="conf", config_name="gpt_sft")
 def main(cfg) -> None:
+    cfg.model = load_and_override_model_config(cfg.model.restore_from_path, cfg.model)
```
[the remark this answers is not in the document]
gshennvm: let's switch to this actually. i've been wanting to do it. ok with the potential breakage. they are easy to fix

examples/nlp/gpt/conf/gpt_dpo.yaml
```diff
@@ -57,6 +57,7 @@ model:
   micro_batch_size: 1
   global_batch_size: 64
   megatron_amp_O2: True
+  variable_seq_lengths: ${not:model.data.pad_length_to_multiple_of}
```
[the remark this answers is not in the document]
arendu: no, if your PR does not need it, then i dont as well.

examples/nlp/gpt/train_gpt_sft.py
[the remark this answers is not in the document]
arendu: oh  I thought i deleted `_modify_config` it should be.

nemo_aligner/data/nlp/scripts/undo_special_tokens.py
arendu: agreed!

nemo_aligner/data/nlp/datasets.py
```diff
@@ -344,16 +350,95 @@ def encode(self, text, append_eod=False):
...
+                ...]
+        returns:
+            conversation:  is a string formatted with the chat template
+        """
+        if OmegaConf.select(self.cfg, "data.chat_prompt_tokens") is not None:
+            special_tokens = self.cfg.data.chat_prompt_tokens
+        else:
+            raise RuntimeError(
+                "You don't have a model (model_config.yaml) which has chat_prompt_tokens, are you sure this is a Chat/Instruction model?"
+            )
```
terrykong, line 413:
```suggestion
        if OmegaConf.select(self.cfg, "data.chat_prompt_tokens") is None:
            raise RuntimeError(
                "You don't have a model (model_config.yaml) which has chat_prompt_tokens, are you sure this is a Chat/Instruction model?"
            )
        special_tokens = self.cfg.data.chat_prompt_tokens
```

nemo_aligner/data/nlp/scripts/undo_special_tokens.py
```diff
@@ -0,0 +1,77 @@
...
+        conversation.append({"role": role, "content": content.strip()})
+
+    empty_asst = conversation.pop()
+
+    return conversation, empty_asst
+
+
+if __name__ == "__main__":
+
+    input_jsonl = sys.argv[1]
```
terrykong, line 45: Could you use argparse, even for a single argument?

nemo_aligner/data/nlp/scripts/undo_special_tokens.py
```diff
@@ -0,0 +1,77 @@
...
...or handling of large input datasets\n -  Not as accurate as transformers\n -  Not as flexible as transformers\n \n\n Pros of transformers:\n \n\n -  Better parallelism and handling of large input datasets\n -  More flexible and versatile\n -  Higher accuracy\n -  Good for actual applications in industry and the real world\n \n\n Cons of transformers:\n \n\n -  More complex and harder to understand and debug\n -  Slower to run\n -  Not as good for teaching and research\n<extra_id_1>User\nFor a chat-based music recommendation platform, which model is recommended? Please mention any model that works well, including those mentioned above or any others.\n<extra_id_1>Assistant\nFor a chat-based music recommendation platform, I would recommend using a transformer architecture, such as the GPT-4 model. Transformers are better for handling large input datasets, and they are more flexible and versatile than RNNs. They are also more accurate, so they are better for actually recommending music to users in a real-world application. RNNs are better for teaching and research, as they are simpler and easier to understand and debug.\n<extra_id_1>User\ncould you add more?\n<extra_id_1>Assistant\n"""
```
terrykong, line 67: just leaving a note that this is better converted into a unit test

nemo_aligner/data/nlp/datasets.py
```diff
@@ -344,16 +350,95 @@ def encode(self, text, append_eod=False):
...
+            "conversations": [],
+            "mask": "User",
+            "type": "VALUE_TO_TEXT",
+        }
+
+        # Extract the system message
+        for msg in input_list:
+            if msg["role"] == "system":
+                output_dict["system"] = msg["content"]
+                break  # Assuming only one system message
```
terrykong, line 382: (disclaimer: not very familiar with this format)

Q: is this assumption of having only one system prompt defined in the schema of the openai format? It looks like the format allows for multiple system prompts at least according to this question on their blogs: https://community.openai.com/t/multiple-system-messages/295258

Since it's allowed, should we just throw an error as opposed to silent ignoring the rest?

terrykong (changes requested):
TODOs
- [x] compatbility test
- [ ] Stretch (update the [dpo.sh](https://github.com/NVIDIA/NeMo-Aligner/blob/main/tests/functional/dpo.sh) template test script to convert the train data json into this new format)

nemo_aligner/data/nlp/datasets.py
arendu: good point, ill throw in an assert -- we should only have one system prompt.

nemo_aligner/data/nlp/scripts/undo_special_tokens.py
arendu: doing this as part of unit tests

tests/test_datasets.py
```diff
@@ -22,8 +22,16 @@
 from nemo.collections.common.tokenizers.huggingface.auto_tokenizer import AutoTokenizer
 from nemo_aligner.algorithms.dpo import dpo_custom_collate
 from nemo_aligner.data.nlp.builders import build_dataloader, build_train_valid_test_dpo_datasets
+from nemo_aligner.data.nlp.scripts.undo_special_tokens import format_conversation
 from nemo_aligner.utils import parallel_state
 
+try:
```
terrykong, line 28: There's something like pytest.mark.skipif

tests/test_datasets.py
```diff
@@ -136,6 +144,54 @@ def test_dpo_loader(init_model_parallel, make_tmp_jsonl, llama3_tokenizer):
...
+
+        jinja_template = """{% for message in conversation %}{%- if message.role == "system" -%}<extra_id_0>System\n{{ message.content }}\n{% elif message.role == "user" -%}<extra_id_1>User\n{{ message.content }}\n{% elif message.role == "assistant" -%}<extra_id_1>Assistant\n{{ message.content }}\n{% endif %}{% endfor %}"""
+        jinja_template = Template(jinja_template)
+        prompt_str_jinja_rendered = jinja_template.render(conversation=oai_messages_prompt)
+        prompt_str_jinja_rendered = remove_trailing(
+            prompt_str_jinja_rendered, "\n"
+        )  # (@adithyare) jinja will add the ending of message token which we should remove to make a prompt.
+        assert prompt_str == prompt_str_jinja_rendered
+
+    return True
```
terrykong, line 192: can remove

terrykong:
closing in favor of #403

tests/test_datasets.py
arendu: @terrykong  do you mean pytest.mark.skipif?

terrykong (approved):

terrykong (approved):

nemo_aligner/data/nlp/datasets.py
```diff
@@ -344,16 +350,97 @@ def encode(self, text, append_eod=False):
...
+                conversation_entry = {
+                    "from": msg["role"].capitalize(),  # Capitalize 'user' and 'assistant'
+                    "value": msg["content"],
+                    "label": None,
+                }
+                output_dict["conversations"].append(conversation_entry)
+
+        return output_dict
+
+    def convert(self, messages):
```
ccclyu, line 399: Do you think it can support `apply_chat_template` (https://huggingface.co/docs/transformers/main/en/chat_templating) for huggingface tokenizers that are adapted in most open-sourced LLMs?

PR #474, opened 2025-01-03
fix: Generate version number

ko3n1g:
# What does this PR do ?

# Changelog 

# Usage

```python
```

**Pre checks**:

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo_aligner/package_info.py
+++ b/nemo_aligner/package_info.py
@@ -23,7 +23,13 @@ DEV = "dev0"
 VERSION = (MAJOR, MINOR, PATCH, PRE_RELEASE, DEV)
 
 __shortversion__ = ".".join(map(str, VERSION[:3]))
-__version__ = __shortversion__ + VERSION[3] + "." + ".".join(VERSION[4:])
+__version__ = __shortversion__
+
+if VERSION[3] != "":
+    __version__ = __version__ + VERSION[3]
+
+if VERSION[4] != "":
+    __version__ = __version__ + "." + ".".join(VERSION[4:])
 
 __package_name__ = "nemo_aligner"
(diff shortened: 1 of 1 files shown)


nemo_aligner/package_info.py
```diff
@@ -23,7 +23,13 @@
 VERSION = (MAJOR, MINOR, PATCH, PRE_RELEASE, DEV)
 
 __shortversion__ = ".".join(map(str, VERSION[:3]))
-__version__ = __shortversion__ + VERSION[3] + "." + ".".join(VERSION[4:])
+__version__ = __shortversion__
```
terrykong, line 26: WDYT about?
```suggestion
__version__ = ".".join(filter(None, map(str, VERSION)))
```
it's a little more succinct functionally

terrykong (changes requested):

nemo_aligner/package_info.py
ko3n1g: Hm, the issue is that pre-release is joined w/o a dit, but the rest is. That makes it so clunky

terrykong: i missed that. good point

nemo_aligner/package_info.py
```diff
@@ -23,7 +23,13 @@
...
+
+if VERSION[3] != "":
+    __version__ = __version__ + VERSION[3]
+
+if VERSION[4] != "":
+    __version__ = __version__ + "." + ".".join(VERSION[4:])
```
terrykong, line 32:
```suggestion
    __version__ = __version__ + "." + ".".join(VERSION[4])
```

terrykong (approved):

PR #485, opened 2025-01-17
fix: pynvml is pinned when using TRTLLM v13 due to breaking change in 12.0.0

terrykong:
# What does this PR do ?

# Changelog 

# Usage

```python
```

**Pre checks**:

# Additional Information
* Related to # (issue)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,7 +38,7 @@ git pull --rebase || true
 pip install --no-cache-dir --no-deps -e .
 EOF
 
-FROM ${BASE_IMAGE} as final
+FROM ${BASE_IMAGE} AS final
 LABEL "nemo.library"="nemo-aligner"
 WORKDIR /opt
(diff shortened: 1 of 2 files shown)


ko3n1g (commented):
What was the reason for placing this dependency here instead into the requirements.txt?

terrykong:
> What was the reason for placing this dependency here instead into the requirements.txt?

At the moment, we install --no-deps, so it actually wouldn't make a difference, but that's a good point that we can also document this in requirements.txt since we'll eventually rely on that

ko3n1g (dismissed):
Thanks for the reminder! SG, I think we’ll anyway be revisiting the dependency management soon

terrykong:
@ko3n1g Updated requirements.txt + comment

ko3n1g (approved):
Thanks for this change!

PR #512, opened 2025-02-18
fix: Fixes for mcore distributed optimizer

ashors1:
# What does this PR do ?

# Changelog 

# Usage

```python
```

**Pre checks**:

# Additional Information
* Related to # (issue)

### DIFF
--- a/nemo_aligner/utils/train_utils.py
+++ b/nemo_aligner/utils/train_utils.py
@@ -77,6 +77,10 @@ def prepare_for_training_step(ptl_model, zero_grad=True):
         if ptl_model.prev_global_batch_size != current_global_batch_size and ptl_model.prev_global_batch_size:
             ptl_model.trainer.should_stop = True
 
+    if ptl_model.use_mcore_dist_optim:
+        for model_chunk in ptl_model.model:
+            model_chunk.zero_grad_buffer()
+
     if zero_grad:
         # we zero grads here because we also call backward in the megatron-core fwd/bwd functions
         ptl_model._optimizer.zero_grad()
--- a/nemo_aligner/utils/utils.py
+++ b/nemo_aligner/utils/utils.py
@@ -564,7 +564,7 @@ def swap_dict(resident_model, cpu_weights, offload_onto_cpu=True, megatron_amp_O
     load_to = resident_model.model
     if isinstance(load_to, list):
         assert (
-            len(model) == 1
+            len(load_to) == 1
         ), "NeMo-Aligner is currently incompatible with virtual pipeline parallel. Please disable VPP."
         load_to = load_to[0]
 
--- a/nemo_aligner/utils/utils.py
+++ b/nemo_aligner/utils/utils.py
@@ -40,6 +40,8 @@ from nemo.utils import AppState, logging
 from nemo.utils.exp_manager import NeMoModelCheckpoint
 from nemo_aligner.models.nlp.gpt.gpt_reward_model import GPTRewardModel
 
+from megatron.core.distributed import DistributedDataParallel
+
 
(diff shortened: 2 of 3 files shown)


nemo_aligner/utils/utils.py
```diff
@@ -500,6 +501,11 @@ def retrieve_model_state_dict_in_cpu(model, megatron_amp_O2=True):
     """
     cpu_dict = {}
 
+    if hasattr(model, "model"):
+        model = model.model
+    if isinstance(model, list):
+        model = model[0]
```
terrykong, line 507: should this assert that the length is 1? might be surprising behavior to drop the other model

nemo_aligner/utils/train_utils.py
```diff
@@ -77,6 +77,10 @@ def prepare_for_training_step(ptl_model, zero_grad=True):
         if ptl_model.prev_global_batch_size != current_global_batch_size and ptl_model.prev_global_batch_size:
             ptl_model.trainer.should_stop = True
 
+    if ptl_model.use_mcore_dist_optim:
```
terrykong, line 80: does this need to be moved under the `if zero_grad:` line?

nemo_aligner/utils/utils.py
```diff
@@ -552,14 +558,19 @@ def swap_dict(resident_model, cpu_weights, offload_onto_cpu=True, megatron_amp_O
     if offload_onto_cpu:
         offloaded_weights = retrieve_model_state_dict_in_cpu(resident_model, megatron_amp_O2=megatron_amp_O2)
 
-    resident_model.load_state_dict(cpu_weights)
+    load_to = resident_model.model
+    if isinstance(load_to, list):
+        load_to = load_to[0]
```
terrykong, line 563: same Q: about the length

for my understanding, is this a change needed b/c of mcore dist opt?

nemo_aligner/utils/train_utils.py
ashors1: This needs to be done regardless of whether `zero_grad` is true. The reason `zero_grad` is sometimes false is because zeroing the gradient is handled from the custom trainer. But I thought it made more sense to zero the grad buffers here to avoid duplicating that code across all trainers.

nemo_aligner/utils/utils.py
ashors1: Yeah, when using mcore optimizer, the model ends up being a list of McoreDDP-wrapped model chunks (references: https://github.com/NVIDIA/NeMo/blob/0dd33007e6e37bc8fb5bff152595e46a82eccdae/nemo/collections/nlp/parts/nlp_overrides.py#L282-L283, https://github.com/NVIDIA/NeMo/blob/main/nemo/collections/nlp/models/language_modeling/megatron_gpt_model.py#L609-L620). This isn't the case when using apex.

nemo_aligner/utils/utils.py
```diff
@@ -552,14 +561,22 @@ def swap_dict(resident_model, cpu_weights, offload_onto_cpu=True, megatron_amp_O
     if offload_onto_cpu:
         offloaded_weights = retrieve_model_state_dict_in_cpu(resident_model, megatron_amp_O2=megatron_amp_O2)
 
-    resident_model.load_state_dict(cpu_weights)
+    load_to = resident_model.model
+    if isinstance(load_to, list):
+        assert (
+            len(model) == 1
```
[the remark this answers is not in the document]
ashors1: Oh yes, thanks for the catch!

terrykong:
Latest CI passed: https://github.com/NVIDIA/NeMo-Aligner/actions/runs/13416682771

recent CI failure due to cherry-picking label

